PR #2, opened 2015-01-27
Feature/secure login cso

dp6ai:


app/controllers/base_controller.rb
```diff
@@ -0,0 +1,12 @@
+class BaseController < ApplicationController
+  include Pundit
+
+  rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
+  before_action { |controller| authorize controller }
+  after_action :verify_authorized
+
+  private
+  def user_not_authorized
+    render text: t('not_authorized'), status: 403
```
ansonK: Shouldn't the status here be 401 (Unauthorized) and not 403 (Forbidden) ?

ansonK: don't listen to me it is right

pixeltrix: Why are we rendering text and not letting rails render a error page?

spec/policies/defence_request_controller_policy_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe DefenceRequestControllerPolicy do
+  subject { DefenceRequestControllerPolicy }
```
pixeltrix: You could right this as `subject { described_class }`. Also you could add a custom example group which sets this by default if there are going to be multiple policies.

spec/policies/defence_request_controller_policy_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  [:index?].each do |action|
+    permissions action do
+
+      it "grants access to #{action} if user role is 'cso'" do
+        expect(subject).to permit(User.new(role: :cso), DefenceRequestControllerPolicy.new)
```
pixeltrix: It the second argument right? From the docs it's meant to be an model or something that can have an action performed on it.

spec/features/user_authentication_feature_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    context 'unauthenticated' do
+      scenario 'logs in with valid credentials' do
+        visit root_path
+        expect(page).to have_button('Sign in')
+      end
+    end
+  end
+end
```
pixeltrix: This example could be made a bit more subjectively readable like this:

``` ruby
RSpec.feature 'user authentication' do
  background do
    User.create!(role: :cso, email: 'cso@example.com', password: 'password')
  end

  context 'when unauthenticated' do
    scenario 'visiting the root url' do
      visit root_path
      expect(page).to have_button('Sign in')
    end
  end

  context 'when signing in' do
    context 'with valid credentials' do
      scenario 'the cso logs in' do
        sign_in 'cso@example.com', 'password'
        expect(page).to have_content('CSO DASHBOARD')
      end
    end

    context 'with invalid credentials' do
      scenario 'logs in with invalid credentials' do
        sign_in 'cso@example.com', 'notapassword'
        expect(page).to_not have_content('CSO DASHBOARD')
      end
    end

    def sign_in(email, password)
      visit new_user_session_path
      fill_in 'user_email', with: email
      fill_in 'user_password', with: password
      click_button 'Sign in'
    end 
  end
end
```

Also adding FactoryGirl allows you to use traits to create your models cleanly, e.g:

``` ruby
FactoryGirl.create :user, :cso, email: cso_email, password: cso_password
```

One more thing - doesn't `create` fail because of the lack of a password confirmation?

db/migrate/20150126150913_devise_create_users.rb
```diff
@@ -0,0 +1,42 @@
...
+      ## Confirmable
+      # t.string   :confirmation_token
+      # t.datetime :confirmed_at
+      # t.datetime :confirmation_sent_at
+      # t.string   :unconfirmed_email # Only if using reconfirmable
+
+      ## Lockable
+      # t.integer  :failed_attempts, default: 0, null: false # Only if lock strategy is :failed_attempts
+      # t.string   :unlock_token # Only if unlock strategy is :email or :both
+      # t.datetime :locked_at
```
pixeltrix: Weren't we having lockable enabled?

PR #5, opened 2015-01-29
Feature/search for own solicitor

FriedSock:


app/controllers/defence_requests_controller.rb
```diff
@@ -5,6 +5,13 @@ def index
 
   def new
     @defence_request = DefenceRequest.new
+    @solicitor = DefenceRequest.new
+  end
+
+  def solicitors_search
+    query_string = params[:q]
+    url = URI.parse "#{Settings.solicitor_search_domain}/solicitors/search/?q=#{query_string}"
+    @solicitors = JSON.parse(HTTParty.post(url).body)['solicitors']
```
ansonK: Were we not using CORS in the solicitor-search so we didn't need to do this sort of proxying?
Or is this just an 'alpha get it done' way?

PR #6, opened 2015-01-29
Feature/cso creates a new dr

dp6ai:


app/controllers/defence_requests_controller.rb
```diff
@@ -14,4 +15,47 @@ def solicitors_search
...
+      params[:defence_request]['time_of_arrival(1i)'] +
+        params[:defence_request]['time_of_arrival(2i)'] +
+        params[:defence_request]['time_of_arrival(3i)'] + ' ' +
+        params[:defence_request]['time_of_arrival(4i)'] + ':' +
+        params[:defence_request]['time_of_arrival(5i)']
+    )
+
+    params[:defence_request]['time_of_arrival'] = time_of_arrival
+
+    params.fetch(:defence_request).permit(:solicitor_type,
```
ansonK: should be params.request as the params should always be nested

dp6ai: good spot. changing

PR #7, opened 2015-01-29
Allow users to change password

dp6ai:
https://www.pivotaltracker.com/story/show/86704942

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,9 +1,9 @@
 Rails.application.routes.draw do
-  devise_for :users
-  # The priority is based upon order of creation: first created -> highest priority.
-  # See how all your routes lay out with "rake routes".
-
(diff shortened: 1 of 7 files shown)


config/routes.rb
```diff
@@ -1,5 +1,9 @@
 Rails.application.routes.draw do
-  devise_for :users
+  devise_for :users, skip: [:registrations]
+  as :user do
+    get 'users/edit' => 'devise/registrations#edit', :as => 'edit_user_registration'
+    patch 'users/:id' => 'devise/registrations#update', :as => 'user_registration'
+  end
   # The priority is based upon order of creation: first created -> highest priority.
```
pixeltrix: Can we delete the example routes - I think we're all capable of reading the docs if necessary

PR #13, opened 2015-02-02
Feature/cso dashboard

dp6ai:


FriedSock:
Maybe we could have a partial for the defence request form that we can reuse between new and edit? We have a lot of upcoming stories involving making little changes to it, and will be a pain to update in 2 places every time.

dp6ai:
Completely agree @FriedSock  it needs a lot mroe tidying i was more wanting feedback on the js

app/views/defence_requests/_table_body.html.erb
```diff
@@ -0,0 +1,20 @@
+<% @defence_requests.each_with_index do |def_req, ind| %>
+  <tr id=<%= "dr_#{def_req.id}" %>>
```
pixeltrix: This will render as `<tr id=dr_100>` - need to change it to this:

``` erb
<tr id="dr_<%= def_req.id %>">
```

alternatively you could use [`content_tag_for`][1]:

``` erb
<% content_tag_for(:tr, def_req) do %>
  <td><%= def_req.solicitor_type %></td>
  …
<% end %>
```

Only problem would be that id would be `defence_request_100` not `dr_100`.

[1]: http://api.rubyonrails.org/classes/ActionView/Helpers/RecordTagHelper.html#method-i-content_tag_for

app/views/defence_requests/edit.html.erb
```diff
@@ -0,0 +1,77 @@
...
+
+        <%= f.check_box :adult, label: t('adult') %>
+
+        <%= f.date_select :date_of_birth, {
+            start_year: Date.today.year,
+            end_year: 120.years.ago.year,
+            label: t('date_of_birth')
+        }, {
+                              class: 'time-select'
+                          } %>
```
pixeltrix: Indenting seems whacked?

PR #14, opened 2015-02-02
Fix bug where DateTime parse error always prevented Defence Request Creation

FriedSock:
Not sure how long this bug has existed for, but the
feature spec was still passing because "Bob Smith"
was part of the params printed in the 500 page for
the error, test is now more robust

### DIFF
--- a/app/controllers/defence_requests_controller.rb
+++ b/app/controllers/defence_requests_controller.rb
@@ -29,16 +29,6 @@ class DefenceRequestsController < BaseController
 
 
   def defence_request_params
-    time_of_arrival = DateTime.parse(
-      params[:defence_request]['time_of_arrival(1i)'] +
-        params[:defence_request]['time_of_arrival(2i)'] +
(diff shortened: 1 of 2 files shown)


pixeltrix:
Is this a Rails bug? Why isn't the standard multi-parameter attributes code handling it correctly?

pixeltrix:
Also shorter commit titles so they don't wrap in the PR form :smile:

FriedSock:
No, the string concat was just missing some dashes so the datestring didn't make any sense, the dashes must have got lost in a merge or something *shrug*. The Github title limit is retarded because you can just go and extend the title, its also completely arbitrary - the title I had was less than 80 characters. Id rather write a descriptive commit message that people will see in git log, gitx etc and just make sure to edit the title on Github if its that bothersome.

pixeltrix:
Still slightly puzzled at why you're doing this in the controller - doesn't the multi-parameter assignment work on the model?

FriedSock:
Oh right, I guess that would be a question for @dp6ai. I can give it a whirl though.

PR #15, opened 2015-02-03
Feature/audit trail

FriedSock:


dp6ai:
Testing? We really need to test we are at least writing audits

FriedSock:
Yep, the would probably be a  good idea.

PR #18, opened 2015-02-04
Close solicitor results

ansonK:
Adds ways of clearing own solicitor search results:
* pressing ESC key
* clicking a Close link in the results

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -48,6 +48,9 @@
 
 .solicitor_results_list{
   padding-left: 14px;
+  position: absolute;
+  margin-top: 60px;
+  z-index: 1;
 }
 
 td.actions{
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -54,3 +54,8 @@ td.actions{
   text-align: center;
 }
 
+[data-fn="btn-solicitor-results-close"] {
+  position: absolute;
+  right: 20px;
+  z-index: 10;
+}
--- a/app/views/defence_requests/new.html.erb
+++ b/app/views/defence_requests/new.html.erb
@@ -1,2 +1,1 @@
 <%= render 'shared/form_template', form_type: 'new', klass: DefenceRequest %>
-
--- a/app/views/defence_requests/_form.html.erb
+++ b/app/views/defence_requests/_form.html.erb
@@ -76,6 +76,14 @@
 <% if @defence_request.new_record? %>
   <script>
     $(document).ready(function () {
+
+      // Clear solicitor search results on ESC keypress
+      $(document).on('keyup', function(event) {
+        if($('.solicitor_results_list *').length > 0 && event.keyCode === 27){
+          $('.solicitor_results_list').empty();
+        }
+      });
+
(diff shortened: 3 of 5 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -48,3 +48,9 @@
 .solicitor_results_list {
   padding-left: 14px;
 }
+
+[data-fn="btn-solicitor-results-close"] {
```
pixeltrix: Is a data attribute selector going to work in IE7? I don't know - just asking the question.

app/views/defence_requests/new.html.erb
```diff
@@ -82,6 +82,13 @@
 
 <script>
  $(document).ready(function () {
+
+  $(document).on('keydown', function(event) {
+    if($('.solicitor_results_list *').length > 0 && event.keyCode === 27){
```
pixeltrix: Not sure whether it'll be a problem but sometime different keyboards report different values: http://unixpapa.com/js/key.html

I think `ESC` is consistent though so it should be okay

spec/features/defence_request_creation_spec.rb
```diff
@@ -91,6 +91,36 @@
...
+     visit root_path
+     click_link 'New Defence Request'
+     choose 'Own'
+     fill_in 'q', with: "Bob Smith"
+
+     click_button 'Search'
+     expect(page).to have_content 'Bobson Smith'
+
+     within('.solicitor_results_list') do
+      click_link("Close")
```
pixeltrix: Just nitpicking but missing a space on the indent

spec/features/defence_request_creation_spec.rb
```diff
@@ -91,6 +91,36 @@
...
+     end
+     expect(page).not_to have_content 'Bobson Smith'
+  end
+
+  scenario "pressing ESC clears solicitor search results", js: true do
+    stub_solicitor_search_for_bob_smith
+     visit root_path
```
pixeltrix: Again just nitpicking but indents messed up

spec/features/defence_request_creation_spec.rb
ansonK: yeah not sure whose editor is not set to 2 spaces, was something I was going to raise

app/assets/stylesheets/application.scss
ansonK: according to https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors IE7 does support the [name=value] selector so should be ok

app/views/defence_requests/new.html.erb
ansonK: afaik ESC is consistent. If there is an issue with it there is still the Close link as a fallback
I'm going to look into something like browserstack for beta so we can be more certain about these sorts of things

app/views/defence_requests/new.html.erb
```diff
@@ -1,2 +1 @@
-<%= render 'shared/form_template', form_type: 'new', object_name: 'defence_request' %>
-
+<%= render 'shared/form_template', form_type: 'new', object_name: 'defence_request' %>
```
FriedSock: We refactored the form template so you will need to pass in `klass: DefenceRequest` after rebasing.

PR #19, opened 2015-02-04
Feature/closing defence request

FriedSock:


FriedSock:
I don't think we should add the entire set of state transitions in this branch, just add them as we go along.

app/controllers/defence_requests_controller.rb
```diff
@@ -24,12 +26,21 @@ def solicitors_search
   def create
     @defence_request = DefenceRequest.new(defence_request_params)
     if @defence_request.save
-      redirect_to({ action: :index }, notice: t('models.create', model: @defence_request.class))
+      redirect_to({ action: :index }, notice: flash_message(:create, DefenceRequest))
```
pixeltrix: Using a hash here is less performant that using a named route helper, e.g:

```
redirect_to defence_requests_url, notice: flash_message(:create, DefenceRequest)
```

app/controllers/defence_requests_controller.rb
```diff
@@ -24,12 +26,21 @@ def solicitors_search
...
       render :new
     end
   end
 
+  def close
+    @defence_request.close
+    if @defence_request.save
+      redirect_to({ action: :index }, notice: flash_message(:close, DefenceRequest))
+    else
+      redirect_to({ action: :index }, notice: flash_message(:failed_close, DefenceRequest))
```
pixeltrix: Same here about using hashes for urls as above

app/views/defence_requests/index.html.erb
```diff
@@ -53,12 +53,12 @@
         <td><%= def_req.detainee_first_name %></td>
         <td><%= def_req.gender %></td>
         <td><%= def_req.adult %></td>
-        <td><%= def_req.date_of_birth %></td>
         <td><%= def_req.appropriate_adult %></td>
         <td><%= def_req.custody_number %></td>
         <td><%= def_req.allegations %></td>
         <td><%= def_req.comments %></td>
         <td><%= def_req.time_of_arrival %></td>
+        <td><%= render partial: 'actions', locals: { def_req: def_req } %></td>
```
pixeltrix: You can omit the `:partial` and `:locals` keys as they are the defaults now, e.g:

``` erb
<%= render 'actions', def_req: def_req %>
```

config/routes.rb
```diff
@@ -10,6 +10,9 @@
     collection do
       post 'solicitors_search'
     end
+    member do
+      put  'close'
+    end
```
pixeltrix: You can write the scope for these actions inline:

``` ruby
post 'solicitors_search', on: :collection
put 'close', on: :member
```

db/migrate/20150204103835_add_state_to_defence_requests.rb
```diff
@@ -0,0 +1,5 @@
+class AddStateToDefenceRequests < ActiveRecord::Migration
+  def change
+    add_column :defence_requests, :state, :string
```
pixeltrix: Do we want the initial state as the column default? I guess by leaving it as `NULL` in the database it only needs changing in the one place so I can see the argument for it - your call.

spec/features/cso_dashboard_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  scenario 'Closing a case' do
+    visit defence_requests_path
+    expect(page).to have_content 'Bob Smith'
+    click_button 'Close'
+
+    expect(page).to have_content "Defence Request successfully closed"
+    expect(page).to_not have_content "Bob Smith"
+  end
+
+end
```
pixeltrix: Nice feature :+1:

spec/models/defence_request_spec.rb
```diff
@@ -1,8 +1,9 @@
 require 'rails_helper'
 
 RSpec.describe DefenceRequest, type: :model do
+  subject { DefenceRequest.new }
```
pixeltrix: The default is `subject { described_class.new }` so there's no need for this.

spec/support/helpers.rb
```diff
@@ -8,4 +8,34 @@ def create_cso_and_login
...
+        check 'defence_request[adult]'
+        select('1976', from: 'defence_request_date_of_birth_1i')
+        select('January', from: 'defence_request_date_of_birth_2i')
+        select('1', from: 'defence_request_date_of_birth_3i')
+        check 'defence_request[appropriate_adult]'
+      end
+      fill_in 'Comments', with: 'This is a very bad man. Send him down...'
+      click_button 'Create Defence Request'
+    end
+ end
```
pixeltrix: Nitpick alert: indenting appears to be off

spec/models/defence_request_spec.rb
```diff
@@ -18,4 +19,11 @@
     it { expect(subject).to allow_value('020 1234 5678', '+44(0)7891234567').for(:phone_number) }
     it { expect(subject).to_not allow_value('1234', 'ABC').for(:phone_number) }
   end
+
+  describe :states do
+    specify { expect(subject.state).to eql 'open' }
+
+    specify { expect{ subject.close }.to_not raise_error }
+    specify { expect{ subject.close; subject.close }.to raise_error }
```
pixeltrix: Do we want to specify at least an error class here, otherwise it could pass when something is wrong

db/migrate/20150204103835_add_state_to_defence_requests.rb
FriedSock: The initial state defaults to `:created `otherwise

config/routes.rb
FriedSock: There are going to be a bunch of different actions on defence requests, so it makes sense to use a block I think?

db/migrate/20150204103835_add_state_to_defence_requests.rb
FriedSock: Yeah the simple_state gem always sets  the field to the default specified in the model, can't really the benefits of having a default at the database level.

spec/models/defence_request_spec.rb
FriedSock: But if you don't explicitly set subject, then it will be set to the default for any nested described blocks, so we surely need this?

pixeltrix: You can override it where it's needed - don't see anything in this spec that needs it.

FriedSock: If I remove the settng of subject then `:validations` becomes the default subject of its block

PR #21, opened 2015-02-04
Edit a DR

dp6ai:
- added a update and edit actions
- partails for the form and shared with new
- policies for cso

https://www.pivotaltracker.com/story/show/86902518


app/assets/stylesheets/application.scss
```diff
@@ -48,3 +48,7 @@
 .solicitor_results_list {
   padding-left: 14px;
 }
+td.actions{
```
pixeltrix: Space before `{` - if I'm nitpicking Anson's and Jack's PRs then I have to do everyone's :smile:

app/controllers/defence_requests_controller.rb
```diff
@@ -30,6 +32,17 @@ def create
     end
   end
 
+  def edit
+  end
+
+  def update
+    if @defence_request.update_attributes(defence_request_params)
+      redirect_to({ action: :index }, notice: t('models.update', model: @defence_request.class))
```
pixeltrix: As I told Jack use a url helper here rather than a hash as it's quicker:

``` ruby
redirect_to defence_requests_url, notice: t('models.update', model: @defence_request.class)
```

app/assets/stylesheets/application.scss
dp6ai: fair enough...

app/views/defence_requests/_form.html.erb
```diff
@@ -0,0 +1,119 @@
+<% if @defence_request.new_record? %>
+  <%= render partial: 'solicitor_search', locals: {object: @defence_request} %>
```
pixeltrix: Should be able to remove the `:partial` and `:locals` keys since these are the defaults since Rails 3, e.g:

``` erb
<%= render 'solicitor_search', object: @defence_request %>
```

app/views/defence_requests/_form.html.erb
```diff
@@ -0,0 +1,119 @@
...
+<% end %>
+
+<%= bootstrap_form_for(@defence_request) do |f| %>
+  <%= render partial: 'shared/form_errors', locals: {object: @defence_request} %>
```
pixeltrix: Ditto

spec/features/defence_request_management_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+
+        within '.edit_defence_request' do
+          within '.details' do
+            fill_in 'Solicitor Name', with: 'Dave Smith'
+            fill_in 'Solicitor Firm', with: 'Broken Solicitors'
+            fill_in 'Phone Number', with: '0207 284 9999'
+            fill_in 'Custody Number', with: '#CUST-9876'
+            fill_in 'Allegations', with: 'BadMurder'
+            select('10', from: 'defence_request_time_of_arrival_4i')
+            select('00', from: 'defence_request_time_of_arrival_5i')
```
pixeltrix: Can we wrap some of this up in a support method - it seems like something that might be used again

spec/features/defence_request_management_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+          end
+
+          within '.detainee' do
```
pixeltrix: Do we need nested `within` blocks - makes the tests dependent on the HTML design

app/views/defence_requests/_form.html.erb
dp6ai: good to know!

spec/features/defence_request_management_spec.rb
dp6ai: I think if its used again that will be the point to create a helper method

spec/features/defence_request_management_spec.rb
dp6ai: good point

PR #22, opened 2015-02-04
Removed the validation on solictor name and firm

dp6ai:
- altered the min length of the detainee name to 2 chars

https://www.pivotaltracker.com/story/show/86712262

### DIFF
--- a/app/models/defence_request.rb
+++ b/app/models/defence_request.rb
@@ -2,12 +2,10 @@ class DefenceRequest < ActiveRecord::Base
 
   phony_normalize :phone_number, default_country_code: 'GB'
 
-  validates :solicitor_name,
-            :solicitor_firm,
-            :detainee_surname,
+  validates :detainee_surname,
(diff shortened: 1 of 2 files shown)


app/models/defence_request.rb
```diff
@@ -2,12 +2,10 @@ class DefenceRequest < ActiveRecord::Base
...
   phony_normalize :phone_number, default_country_code: 'GB'
 
-  validates :solicitor_name,
-            :solicitor_firm,
-            :detainee_surname,
+  validates :detainee_surname,
             :detainee_first_name,
             :allegations,
-            length: { minimum: 5 }
+            length: { minimum: 2 }
```
pixeltrix: There are [single character chinese surnames][1] though I suspect they'll be entered as the romanized version since I doubt the average custody officer's ability to enter Mandarin.

[1]: http://en.wikipedia.org/wiki/Chinese_surname

PR #23, opened 2015-02-05
Alter DefenceRequest validations and add seeds

dp6ai:
- remove full phone number validations
- change detainee to just save one name
- alter validations to only check presence
- add test users seed data
- added heroku schedule to reset test DB

https://www.pivotaltracker.com/story/show/87704886


db/migrate/20150128172027_add_attributes_to_defence_request.rb
```diff
@@ -4,8 +4,7 @@ def change
     add_column :defence_requests, :solicitor_firm, :string
     add_column :defence_requests, :scheme, :string
     add_column :defence_requests, :phone_number, :string
-    add_column :defence_requests, :detainee_surname, :string
-    add_column :defence_requests, :detainee_first_name, :string
+    add_column :defence_requests, :detainee_name, :string
```
ansonK: Is this changing a migration that has already been deployed rather than creating a new migration?

PR #24, opened 2015-02-06
Feature/solicitor login

FriedSock:


FriedSock:
Got some overlap with 2 different dashboard specs, I think I will make a cleanup branch instead of dealing with that in this PR, as I think the duplication appeared previously.

app/controllers/base_controller.rb
```diff
@@ -2,11 +2,11 @@ class BaseController < ApplicationController
   include Pundit
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
-  before_action { |controller| authorize controller }
-  after_action :verify_authorized
```
dp6ai: i think this should stay in, as discussed

PR #25, opened 2015-02-06
Feature/dscc new dr notify

dp6ai:


app/controllers/defence_requests_controller.rb
```diff
@@ -3,7 +3,8 @@ class DefenceRequestsController < BaseController
   before_action :find_defence_request, only: [:edit, :update, :close]
 
   def index
-    @defence_requests = DefenceRequest.open.order(created_at: :asc)
+    @open_requests = DefenceRequest.open.order(created_at: :asc)
+    @new_requests = DefenceRequest.created.order(created_at: :asc)
   end
```
FriedSock: Heads up: this will need to change in rebase to use policy_scope

spec/models/defence_request_spec.rb
```diff
@@ -15,9 +15,36 @@
...
+      expect{ subject.finish }.to raise_error(Transitions::InvalidTransition)
+      subject.update_current_state(:open)
+      expect{ subject.finish }.to_not raise_error
+      expect(subject.current_state).to eql :finished
+      expect(subject.can_transition? :open).to eq false
+      expect(subject.can_transition? :close).to eq false
+      expect(subject.can_transition? :created).to eq false
+      expect(subject.can_transition? :finish).to eq false
+    end
   end
```
FriedSock: I can see this being a pain to update as the state model grows, we could loop through all the states that we don't expect to transition to instead of copying and pasting the assertions.

spec/support/helpers.rb
```diff
@@ -1,10 +1,10 @@
...
+  def create_role_and_login(role)
     password = '123456789'
-    cso_user = User.create(email: 'cso@example.com', password: password, role: :cso)
+    user = User.create(email: 'cso@example.com', password: password, role: role.to_sym)
     visit new_user_session_path
-    fill_in 'user_email', with: cso_user.email
+    fill_in 'user_email', with: user.email
     fill_in 'user_password', with: password
     click_button 'Sign in'
   end
```
FriedSock: Lovely change :whale2:

PR #27, opened 2015-02-09
Feature/dscc number

FriedSock:


app/views/defence_requests/_form.html.erb
```diff
@@ -62,6 +62,15 @@
     </div>
   </div>
 
+    <div class="panel panel-default DSCC number">
+      <div class="panel-heading">
+        <h3 class="panel-title"><%= t('dscc_number') %></h3>
+      </div>
+      <div class="panel-body">
+        <%= f.text_field :dscc_number, label: t('dscc_number'), readonly: !current_user.cco? %>
```
ansonK: we should be using Policies to control authorization so all authorization is controlled in 1 place and not scattered across the app

https://github.com/elabs/pundit#headless-policies

PR #28, opened 2015-02-09
Only open defence requests if they can transition

FriedSock:
- Also added in a policy for opening DRs to keep
consistent with our guarding of actions.

### DIFF
--- a/app/controllers/defence_requests_controller.rb
+++ b/app/controllers/defence_requests_controller.rb
@@ -3,7 +3,7 @@ class DefenceRequestsController < BaseController
   before_action :find_defence_request, only: [:edit, :update, :close]
 
   def index
-    @open_requests = policy_scope(DefenceRequest).open.order(created_at: :asc)
+    @open_requests = policy_scope(DefenceRequest).opened.order(created_at: :asc)
     @new_requests = policy_scope(DefenceRequest).created.order(created_at: :asc)
   end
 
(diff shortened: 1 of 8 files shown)


app/controllers/defence_requests_controller.rb
```diff
@@ -3,7 +3,7 @@ class DefenceRequestsController < BaseController
   before_action :find_defence_request, only: [:edit, :update, :close]
 
   def index
-    @open_requests = policy_scope(DefenceRequest).open.order(created_at: :asc)
+    @open_requests = policy_scope(DefenceRequest).opened.order(created_at: :asc)
```
dp6ai: :+1:

app/controllers/defence_requests_controller.rb
```diff
@@ -37,7 +37,7 @@ def edit
   end
 
   def update
-    @defence_request.open if current_user.cco?
+    @defence_request.open if policy(@defence_request).open?
```
dp6ai: :+1:

app/policies/defence_request_policy.rb
```diff
@@ -54,4 +54,8 @@ def dscc_number_edit?
     user.cco?
   end
 
+  def open?
+    user.cco? && record.can_transition?(:open)
+  end
```
dp6ai: :+1:

PR #39, opened 2015-02-16
Feature/mark dr accepted

dp6ai:


config/application.rb
```diff
@@ -46,5 +46,6 @@ class Application < Rails::Application
     # Feedback URL (URL for feedback link in phase banner)
     # Use 'auto_add_path' for it to add a path link to the new_feedback route
     config.feedback_url = config.relative_url_root + '/feedback/new'
+    config.action_mailer.default_url_options = { host: 'example.com' } #TODO: this needs correcting
```
ansonK: shouldn't this be an ENV variable?

dp6ai: It can be. I wanted to add a TODO and the settings.yml would have been unhappy. It can be fixed later i think

FriedSock:
Quite a lot of logic in the controller, we could ideally add some fancy validations to the model instead but I don't think it is worth the time to refactor at this point if we are "rebuilding" in a few days.

dp6ai:
@FriedSock completely agree, the refactor can wait until after alpha

PR #43, opened 2015-02-18
Feature/resend details

FriedSock:


ansonK:
Does not pass 2 A/C:
* No confirmation before sending
* No DSCC number in notification email

ansonK:
Only tests that clicking the "resend details" btn displays a flash message.
Does not test that an email is sent

PR #45, opened 2015-02-19
Associate a CCO with a DR on opening

dp6ai:
- add association
- alter policy

### DIFF
--- a/app/policies/defence_request_policy.rb
+++ b/app/policies/defence_request_policy.rb
@@ -35,7 +35,7 @@ class DefenceRequestPolicy < ApplicationPolicy
   end
 
   def edit?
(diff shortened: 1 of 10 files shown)


app/policies/defence_request_policy.rb
```diff
@@ -35,7 +35,7 @@ def create?
   end
 
   def edit?
-    (user.cso? && record.created?) || (user.cco? && record.opened?)
+    (user.cso? && record.created?) || (user.cco? && record.opened? && record.cco == user)
```
ansonK: is having ```user.cco? && record.cco == user``` redundant here?
just check ```user.cco == user```

dp6ai: :+1:

PR #47, opened 2015-02-19
Feature/add time of arrival

FriedSock:


ansonK:
this is wrong - the solicitor should *not* be able to change the interview start time... Only the cso should be able to do that.

This story needs to add a new field on the DR - solicitor_estimated_arrival_time or similar

PR #56, opened 2015-02-26
Fix missing i18n translations

dp6ai:
- added missing i18
- moved 'content' files to locale specific versions

https://www.pivotaltracker.com/story/show/88618676


FriedSock:
Seems fine apart from all the simplecov stuff

PR #65, opened 2015-03-02
Refactor policy and transitions

FriedSock:
Previously we had to duplicate logic in the policy
because `can_transition` does not respect guards.

Also remove unused `accepted?` method.


ansonK:
lgtm

PR #71, opened 2015-03-03
Feature/text fields for dates with val

dp6ai:


ansonK:
lgtm
not entirely comfortable with this but its working, lets get it merged in and come back to it later

PR #87, opened 2015-03-16
[logstash_format_logging] Added lograge and logstash-event in production

oskarpearson:


ansonK:
Looks good.
How feasible is it to also add to each log line?: (maybe using a TaggedLogger ?)
- UUID (request id) http://api.rubyonrails.org/classes/ActionDispatch/RequestId.html
- User ID (if there is a current user)

oskarpearson:
I'm going to merge this initial version, and add the more detailed logging (with request ids) in a separate PR, based on the more detailed story at https://www.pivotaltracker.com/story/show/90504218

PR #88, opened 2015-03-16
[unify_environment_settings] Cleaned handling of environment variables

oskarpearson:
(Note that everyone will probably need to do a 'mv .env .env.local' before pulling this, given the way that dotenv 2 works.)

* Changed so that we use ENV.fetch instead of ENV[] - this way we receive
  errors if variables are not defined, rather than just undefined behaviour.
* Migrate to Dotenv 2.0.0
* Adjust .env config files to match Dotenv 2.0.0 structure
* Cleaned up naming of some environment variables
* Grouped and cleaned up the config/settings.yml file
* Corrected config of config.action_mailer.default_url_options to use the url
  rather than the mailserver name
* Migrated database.yml to use ENV variables universally
* Added unicorn config using UNICORN_PORT environment variable to match other
  Ministry of Justice projects
* The Settings.dsds.solicitor_search_domain wasn't just a domain. The setting
  included a full url fragment, including https:// or http:// so I renamed it
  Settings.dsds.solicitor_search_url_prefix instead
* Set enable_starttls_auto to true on our mail config, since it's recommended to be
  true by both https://sendgrid.com/docs/Integrate/Frameworks/rubyonrails.html and
  https://devcenter.heroku.com/articles/sendgrid


ansonK:
lgtm tomorrow after user testing sessions

PR #90, opened 2015-03-16
Replace Open state with Acknowledged state

chriscarter90:


oskarpearson:
LGTM - can I check if we need to do a data migration to match? I'm assuming not, since we don't have any real info in the db...

chriscarter90:
Good point, not sure. Probably can't hurt.

ansonK:
Yes please add a data migration to this. Good rule of thumb is any PR that makes changes that relies on a dataset should include a migration so it will work regardless of whether the target environment has data or not

chriscarter90:
@ansonK @oskarpearson Poke. :cactus:

oskarpearson:
Looks good to me.

PR #91, opened 2015-03-17
Raise on unpermitted params across all envs

pedrosmmoreira:
Setting ActionController to raise on unpermitted params can save a lot
of debugging time in test/dev.

It will also be the default setting at some point in the near future.


db/schema.rb
```diff
@@ -54,12 +54,12 @@
...
     t.boolean  "adult"
     t.boolean  "appropriate_adult",         default: false, null: false
+    t.string   "state"
     t.datetime "created_at"
     t.datetime "updated_at"
-    t.string   "state"
     t.string   "dscc_number"
-    t.text     "feedback"
     t.datetime "interview_start_time"
+    t.text     "feedback"
```
pixeltrix: I think these changes are fine now but if we were in production you'd need to be careful about not getting your schema columns out of order with the production database.

pedrosmmoreira: @pixeltrix Since I just pulled the repo and migrated, am I correct in assuming that it is out of sync further upstream?

PR #93, opened 2015-03-23
Implement blank properly on all field objects

FriedSock:
Fixes a bug where forms would fail to submit even
when there were no errors.

I know this isn't very DRY, but it will be in the mysterious `paradats` branch.


chriscarter90:
:+1: :cactus:

PR #95, opened 2015-03-24
Fix queued state

FriedSock:


FriedSock:
This also seems to fix the story about acknowledged state as well https://www.pivotaltracker.com/story/show/90002384

PR #97, opened 2015-03-25
Refactor flash and error messages

FriedSock:


app/helpers/application_helper.rb
```diff
@@ -1,25 +1,30 @@
...
-  def list_errors(object, field_name)
-    if object.errors.any?
-      unless object.errors.messages[field_name].blank?
-        content_tag :ul do
-          object.errors.messages[field_name].collect do |field|
-            concat(content_tag(:li, field))
-          end
-        end
-      end
+  def flash_message key, msg
```
pixeltrix: What's our standard on using parentheses with method definitions?

app/helpers/application_helper.rb
```diff
@@ -1,25 +1,30 @@
...
+    content_tag(:div, class: "#{key}-summary") do
+      content_tag :p, msg
+    end
+  end
+
+  def object_error_messages active_model_messages
+    content_tag(:ul) do
+      active_model_messages.map do |field_name, field_messages|
+        errors_for_field field_name, field_messages
```
pixeltrix: will concating here not get added to the tag?

pixeltrix:
Other than the minor points :+1:

app/helpers/application_helper.rb
ansonK: My preference is to use parentheses for method definitions

chriscarter90:
Crystal likes this. :+1:

PR #101, opened 2015-03-26
[git_deploy] Added git-deploy

oskarpearson:
* Integrated https://github.com/mislav/git-deploy as a temporary solution for deployments
* Make Unicorn socket and pids relative to the unicorn.rb file
* Configured Unicorn - Added PID file
* Configured Unicorn - Added socket with the best-practice backlog of 64, so that
  errors fail-over faster
* Added tmp/pids and tmp/sockets creation to deploy

### DIFF
--- a/config/unicorn.rb
+++ b/config/unicorn.rb
@@ -1,8 +1,8 @@
 worker_processes Integer(ENV.fetch("WEB_CONCURRENCY", 3))
 listen Integer(ENV.fetch('UNICORN_PORT', 3000))
 
-pid File.join(File.dirname(__FILE__), '..', 'tmp', 'pids', 'unicorn.pid')
-listen File.join(File.dirname(__FILE__), '..', 'tmp', 'sockets', 'unicorn.sock'), :backlog => 64
+pid File.expand_path('../../tmp/pids/unicorn.pid', __FILE__)
+listen File.expand_path('../../tmp/sockets/unicorn.sock', __FILE__), backlog: 64
 
 timeout 15
 preload_app true
--- a/config/unicorn.rb
+++ b/config/unicorn.rb
@@ -1,6 +1,9 @@
 worker_processes Integer(ENV.fetch("WEB_CONCURRENCY", 3))
 listen Integer(ENV.fetch('UNICORN_PORT', 3000))
 
+pid File.join(File.dirname(__FILE__), '..', 'tmp', 'pids', 'unicorn.pid')
+listen File.join(File.dirname(__FILE__), '..', 'tmp', 'sockets', 'unicorn.sock'), :backlog => 64
+
 timeout 15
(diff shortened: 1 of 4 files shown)


config/unicorn.rb
```diff
@@ -1,6 +1,9 @@
 worker_processes Integer(ENV.fetch("WEB_CONCURRENCY", 3))
 listen Integer(ENV.fetch('UNICORN_PORT', 3000))
 
+pid File.join(File.dirname(__FILE__), '..', 'tmp', 'pids', 'unicorn.pid')
+listen File.join(File.dirname(__FILE__), '..', 'tmp', 'sockets', 'unicorn.sock'), :backlog => 64
+
```
pixeltrix: Would this be better? 

``` ruby
pid File.expand_path("../../tmp/pids/unicorn.pid", __FILE__)
listen File.expand_path("../../tmp/sockets/unicorn.sock", __FILE__), backlog: 64
```

Also please use the new hash style :smile:

oskarpearson: Thanks for feedback. It certainly looked ugly. Will fix it now.

chriscarter90: :cactus:

deploy/before_restart
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env ruby
+oldrev, newrev = ARGV
+
+def run(cmd)
+  exit($?.exitstatus) unless system "umask 002 && #{cmd}"
+end
+
+RAILS_ENV   = ENV['RAILS_ENV'] || 'production'
```
pixeltrix: I have a concern about doing the fallback to production like this - what if the `RAILS_ENV` is set to an empty string because of an error in a ansible or salt script, e.g:

``` irb
$ RAILS_ENV= irb
>> ENV['RAILS_ENV'] || 'production'
=> ""
```

Not sure what the answer is :smile:

oskarpearson: Hey. The file is autogenerated by https://github.com/mislav/git-deploy - I'd prefer to keep it the same if possible. But I'm happy to fall back to an invalid environment instead if you want :)

deploy/before_restart
```diff
@@ -0,0 +1,39 @@
...
+use_bundler = File.file? 'Gemfile'
+rake_cmd    = use_bundler ? 'bundle exec rake' : 'rake'
+
+if use_bundler
+  bundler_args = ['--deployment']
+  BUNDLE_WITHOUT = ENV['BUNDLE_WITHOUT'] || 'development:test'
```
pixeltrix: is the `:` right here? The [docs][1] seem to suggest space separated.

[1]: http://bundler.io/v1.9/groups.html

chriscarter90:
:+1:

PR #105, opened 2015-04-02
CSO aborts defence request

robmckinnon:
* CSO can put Defence Request into `aborted` state for the call centre to handle offline
* Remove the `closed` state and the `feedback` field from defence requests
* Implements feature: https://www.pivotaltracker.com/story/show/90003414


FriedSock:
Mostly :+1: but no test for when a request can't be aborted. 
screen shot 2015-04-07 at 11 31 17

The horror!

PR #106, opened 2015-04-02
Use login_as() helper to speed up rspec run.

robmckinnon:
* Instead of signing in via login page for each rspec feature, use login_as() helper.
* My local test run time went from 20s down to 12s after this change.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,6 @@ gem "govuk_elements_rails", "~> 0.1.1"
 
 group :development do
   gem "guard-rspec", require: false
-  gem "terminal-notifier-guard"
   gem "web-console", "~> 2.0"
 end
 
@@ -42,6 +41,7 @@ group :development, :test do
   gem "pry-rails"
   gem "quiet_assets", "~> 1.1"
   gem "rspec-rails", "~> 3.2.0"
+  gem 'rb-fsevent' if `uname` =~ /Darwin/
 end
 
 group :test do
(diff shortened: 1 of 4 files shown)


pedrosmmoreira:
While that is a considerable performance boost, I'd be reluctant to use it across all integration tests.

This is equivalent of saying "Test mode, I want to exempt you from the auth mechanism in place", which is fine: in many places, we don't actually care about the auth - just that it is in the state required. However, I would leave a scenario walking through those steps and guarding agains regression. 

Much as speed is great, it is not the only factor at play.

pedrosmmoreira:
While that is a considerable performance boost, I'd be reluctant to use it across all integration tests.

This is equivalent of saying "Test mode, I want to exempt you from the auth mechanism in place", which is fine: in many places, we don't actually care about the auth - just that it is in the state required. However, I would leave a scenario walking through those steps and guarding against regression. 

Much as speed is great, it is not the only factor at play.

robmckinnon:
Ok. If we add a new login feature we can explicitly test those login page steps. For the other features we just need a particular type of user to be logged in, no need to repeat the login page steps each time.

pixeltrix:
Firstly, this has extraneous changes regarding adding rb-fsevent and removing terminal-notifier-guard - these seem orthogonal to using warder test helpers so please keep PRs atomic.

On the main changes this isn't a good idea because testing the login steps alone really amounts to replicating Devise's tests and removing them from the features runs the risks of bugs slipping through. I can think of plenty of situations where changes in an individual controller ends up with authentication bypassed and removing the login steps may prevent the detection of the problems.

The fact that features take 20s instead of 12s is pretty much irrelevant as you won't be running the full suite that often - that's the CI's job.

ansonK:
Agree with the above. A small speed increase is not worth the risks it opens up

PR #110, opened 2015-04-09
Feature/cso cannot edit dr fields

dp6ai:


app/models/defence_request.rb
```diff
@@ -83,7 +83,7 @@ def own_solicitor?
   private
 
   def format_phone_number
-    self.phone_number = self.phone_number.gsub(/\D/, '')
+    self.phone_number = self.phone_number.gsub(/\D/, '') if self.phone_number
```
pixeltrix: Any reason we can't do this in a setter override and lose the callback?

```
def phone_number=(new_value)
  super(format_phone_number(new_value))
end

private

def format_phone_number(phone_number)
  phone_number.to_s.gsub(/\D/, '') if phone_number
end
```

Also are we sure that we want to enforce digits only on phone numbers?

app/views/defence_requests/form_partials/_solicitor_details.html.erb
```diff
@@ -1,11 +1,11 @@
 <div class="form-group">
   <%= f.label :solicitor_name, t('full_name') %>
-  <%= f.text_field :solicitor_name, class: 'text-field text-field-wide' %>
+  <%= f.text_field :solicitor_name, disabled: !@policy.edit_solicitor_details?, class: 'text-field text-field-wide' %>
```
pixeltrix: Seems a bit ugly - shouldn't this kind of thing be handled by the form object?

pixeltrix:
Can we please squash the WIP commits - there's no need for them in the history

app/views/defence_requests/form_partials/_solicitor_details.html.erb
dp6ai: We're using pundit policies everywhere to handle view logic. I don't think this will bring us much other than extra complexity.

ansonK: This form is being redesigned soon so drying up / refactoring can be done then

ansonK:
So looks like Acknowledged DefenceRequests that are Duty cannot have the solicitor fields edited by a CCO .. ?

dp6ai:
@ansonK im not sure this is the behaviour we want but ive changed it back to allow full editing...

PR #113, opened 2015-04-10
[unicorn_config_cleanups_and_unify_gitignore] Unicorn and .gitignore cleanups

oskarpearson:
* Make .gitignore standard across projects
* Make .gitignore support keeping subdirectories in the tmp/ directory
* Change Unicorn to log stderr and stdout into the log directory, unless run from Foreman


chriscarter90:
:+1:

PR #114, opened 2015-04-10
Feature/dr extra details

dp6ai:


config/locales/en.yml
```diff
@@ -9,10 +9,12 @@ en:
   acknowledged_defence_request: "Acknowledged Defence Requests"
   actions: "Actions"
   add_expected_time_of_arrival: "Add Expected Time of Arrival"
+  address_1: Address 2
+  address_2: Address 1
```
ansonK: Address label translations the wrong way round

dp6ai: :(

spec/factories/defence_request.rb
```diff
@@ -1,5 +1,5 @@
 FactoryGirl.define do
-  now = DateTime.current
+  now = DateTime.new(2001,1,1,1,1)
```
pixeltrix: Can we please use `Time` rather than `DateTime` - it doesn't have the concept of named timezones just offsets, e.g:
``` irb
>> Time.current.time_zone
=> (GMT+00:00) Europe/London
>> DateTime.current.time_zone
NoMethodError: undefined method `time_zone' for Mon, 13 Apr 2015 11:11:09 +0100:DateTime

>> Time.current.zone
=> "BST"
>> DateTime.current.zone
=> "+01:00"
```

Only use `DateTime` when you're doing historical date calculations and you need to take account of calendar reforms.

PR #115, opened 2015-04-13
wip: Generate DSCC number

ansonK:
Generates DSCC number automatically for a Defence Request transitioning into acknowledged state

DSCC number cannot be edited or removed via the UI

Each Defence Request gets 1 DSCC number that cannot be changed


pixeltrix:
What's the reason for using a separate AR model and table? Is there a reason that the request `created_at ` doesn't give us what we need for db searchability? Do we need to access the individual fields of the DSCC number or is it just the whole thing? Any reason we can't used [`composed_of`][1] or even override the attribute methods and just have a string column on the request object and use a PORO instead of an AR model?

[1]: http://api.rubyonrails.org/classes/ActiveRecord/Aggregations/ClassMethods.html#method-i-composed_of

app/models/dscc_number.rb
```diff
@@ -0,0 +1,29 @@
+class DsccNumber < ActiveRecord::Base
+
+  def self.current_monthly_number
+    where(year_and_month: current_year_and_month).order('number desc').limit(1).pluck(:number).first || 0
```
pixeltrix: Isn't this a massive race condition? If two requests are generated at the same time then they would end up with the same `current_monthly_number` wouldn't they?

app/models/dscc_number.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def self.next_monthly_number
+    current_monthly_number + 1
+  end
+
+  def self.current_year_and_month
+    Time.current.to_date.beginning_of_month
```
pixeltrix: Since we're never actually storing this, just using it to prefix the number, we probably could just use `Date.current` - don't even need the `beginning_of_month`.

pixeltrix: Scratch that - we are storing it. :smile:

But why are we storing it - does `created_at` not give us what we need?

app/models/dscc_number.rb
```diff
@@ -0,0 +1,29 @@
...
+      dscc_number.number = next_monthly_number
+    end
+  end
+
+  has_one :defence_request
+
+  validates_uniqueness_of :number, scope: [:year_and_month]
+
+  def to_s
+    [year_and_month.strftime("%y%m"), number.to_s.rjust(5, "0"), extension].join ""
```
pixeltrix: Seems a little convoluted - we could just use `%`, e.g;

``` ruby
def to_s
  "%s%05d%s" % [prefix, number, extension]
end

def prefix
  created_at.strftime("%y%m")
end
```

we could even hard code the extension since it's never going to change.

pixeltrix:
Okay, I think I've worked out the reasoning for creating the AR model - it's so you can generate a monthly incrementing number I guess? I'm sure there must be a better way of generating this.

Also one other thing - we have a 5 digit number which puts a hard limit of 99,999 requests each month. Given that [there were 1.2 million arrests in 2012][1] doesn't this seem dangerously close to overflowing?

[1]: https://www.gov.uk/government/publications/police-powers-and-procedures-in-england-and-wales-201112/police-powers-and-procedures-in-england-and-wales-2011-12#key-facts

pixeltrix:
To illustrate how small that figure is, that's one arrest every 30 seconds approximately.

app/models/dscc_number.rb
ansonK: The DSCC number is keyed off when the DefenceRequest is raised, not when the DsccNumber is generated.
If the DefenceRequest is raised just before midnight on the last day of the month then its DsccNumber must be part of the same month. 
Hence storing the year_and_date separately to DsccNumber#created_at

ansonK:
We are using the existing LAA DSCC number format which we cannot change. Fully agree it is limiting.

The format is yymmnnnnnc and the nnnnn is a 5 digit number that begins again every month so storing the number separately makes looking it up easier

app/models/dscc_number.rb
```diff
@@ -0,0 +1,19 @@
+class DsccNumber < ActiveRecord::Base
+  has_one :defence_request
+
+  validates_uniqueness_of :number, scope: [:year_and_month]
```
pixeltrix: Still a race condition - need to have a unique index and then capture the error and retry

oskarpearson:
One problem with splitting the year-and-month and number is that we can't create a foreign key for defence_request.dscc_number -> dscc_numbers.

I know that foreign keys are generally not used (or used very sparingly) in rails, but this concerns me a bit.

app/services/dscc_number_generator.rb
```diff
@@ -0,0 +1,28 @@
...
+  def year_and_month
+    defence_request.created_at.to_date.beginning_of_month
+  end
+
+  def current_monthly_number
+    @current_monthly_number ||= DsccNumber.most_recent_monthly(year_and_month).pluck(:number).first || 0
+  end
+
+  def next_monthly_number
+    current_monthly_number + 1
```
oskarpearson: Unfortunately this code is definitely subject to race conditions as-is.

Creating an unique compound index on the combination of keys and retrying this a few times will work... but overall it's not ideal. We'll have multiple processes at a ruby level re-trying and then locking, with many milliseconds of wait between the retry and the locking.

I'm sure there must be a better way :) Let me think about it a bit.

oskarpearson:
I found this thread, which is relevant: http://www.postgresql.org/message-id/20081118092652.GB19076@a-kretschmer.de

http://www.postgresql.org/docs/9.1/static/functions-sequence.html says:
```
*nextval*
Advance the sequence object to its next value and return that value. This is done atomically: even if multiple sessions execute nextval concurrently, each will safely receive a distinct sequence value.
```

Here's some test code. The only thing I don't really like here is pre-creating 100 years of sequences. It's a once-off process, though, and could be done in seed data.

```
START_DATE = Date.today.beginning_of_month

def sequence_name(year_and_month)
  sprintf "dscc_number_%04d%02d", year_and_month.year, year_and_month.month
end

(100 * 12).times do |elapsed_months|
  year_and_month = START_DATE + elapsed_months.months
  
  ActiveRecord::Base.connection.execute("create sequence #{sequence_name(year_and_month)}")
end

def dscc_number(year_and_month)
  sql = "select nextval('#{sequence_name(year_and_month)}') as dscc_number"
  # puts sql

  next_item = Integer(ActiveRecord::Base.connection.execute(sql)[0]["dscc_number"])
  
  return sprintf("%d%d%05dZ", year_and_month.year, year_and_month.month, next_item)
end

puts "Sequence examples:"
puts "  #{START_DATE} - 1st call #{dscc_number(START_DATE)}"
puts "  #{START_DATE} - 2nd call #{dscc_number(START_DATE)}"

puts "  #{START_DATE + 1.month} - 1st call #{dscc_number(START_DATE + 1.month)}"
puts "  #{START_DATE + 1.month} - 2nd call #{dscc_number(START_DATE + 1.month)}"

puts "  #{START_DATE + 1.year} - 1st call #{dscc_number(START_DATE + 1.year)}"
puts "  #{START_DATE + 1.year} - 2nd call #{dscc_number(START_DATE + 1.year)}"
puts "  #{START_DATE + 1.year} - 3rd call #{dscc_number(START_DATE + 1.year)}"

# Uncomment to drop sequences at the end for testing
# (100 * 12).times do |elapsed_months|
#   year_and_month = START_DATE + elapsed_months.months
#
#   ActiveRecord::Base.connection.execute("drop sequence #{sequence_name(year_and_month)}")
# end
```

pixeltrix:
I think @oskarpearson's suggestion is probably the correct one to take - it's the only one that I've been able to make work reliably with excessive retries, e.g:

``` ruby
require 'active_record'

# Assumes that you have a database called defence_request in postgresql accessible to the default user
ActiveRecord::Base.establish_connection(adapter: 'postgresql', encoding: 'unicode', database: 'defence_request')
ActiveRecord::Base.logger = Logger.new('/dev/null')

ActiveRecord::Schema.define do
  create_table :defence_requests, force: true  do |t|
    t.string  :first_name,  limit: 50
    t.string  :last_name,   limit: 50
    t.string  :dscc_number, limit: 10
    t.timestamps null: false
  end

  add_index :defence_requests, :dscc_number, unique: true

  prefix = Date.current.strftime('%y%m')
  execute "DROP SEQUENCE IF EXISTS dscc_number_#{prefix}_seq"
  execute "CREATE SEQUENCE dscc_number_#{prefix}_seq MINVALUE 1 MAXVALUE 99999 START WITH 1 NO CYCLE"
end

class DefenceRequest < ActiveRecord::Base
  DSCC_SUFFIX = 'Z'

  def generate_dscc_number!
    unless persisted?
      raise RuntimeError, "Defence requests must be saved before generating a DSCC number"
    end

    prefix = created_at.strftime('%y%m')
    index  = next_dscc_number(prefix)

    update!(dscc_number: ('%s%05d%s' % [prefix, index, DSCC_SUFFIX]))
  end

  private

  def next_dscc_number(prefix)
    self.class.connection.select_value("SELECT nextval('dscc_number_#{prefix}_seq')")
  end
end

10.times do
  pid = Process.fork do
    ActiveRecord::Base.connection.reconnect!

    1000.times do
      request = DefenceRequest.create!(first_name: 'Andrew', last_name: 'White')
      request.generate_dscc_number!

      puts "Process: #{Process.pid}, DSCC No: #{request.dscc_number}\n"
    end
  end
end

Process.waitall
```

Running this script gives the following output:

``` bash
$ time ruby dscc_numbers.rb
real	0m15.926s
user	0m39.022s
sys	0m9.041s
```

Given the simplicity of the implementation I think having another model for this is overkill - even putting it in a service/generator that runs via a background job would probably be too much but I'll leave it up to @ansonK to decide about that.

FriedSock:
I agree using `nextval` is probably the best way to implement this. I am going to close this PR as there is a lot of rebasing required due to divergence from master, and changing the existing implementation.

ansonK:
Having some time to think about this we need to limit the creation of DSCC numbers to be 1 at a time. This Queue could be implemented in the application layer but I think it would be more robust implemented in the database layer.
@oskarpearson's approach is along the right lines, but needing to either define a new sequence or reset the current sequence each month is brittle and feels like an accident waiting to happen.

How about using a CTE? That way selecting the next DSCC number can happen in the same scope as the update that sets it in an atomic way with only 1 db query being issued from the application layer.

```
WITH next_dscc_number as (
  ... figure out next number here ...
)

UPDATE defence_requests SET dscc_number = next_dscc_number WHERE id = 1;
```

oskarpearson:
@ansonK If it's anything like the other dbs I've worked on, it's still subject to race conditions. The only difference is that the time window where the race condition could occur would be smaller.

See http://stackoverflow.com/a/15750424 for more information:

```
Yes, this can most certainly create a race condition, because while all statements are guaranteed atomic, this does not require them to have operated across an unchanging data set during the separate parts of the query's execution.
```

There are some proposed solutions on that page. However they effectively involve locking the full table, which creates a point of lock contention. The sequence stuff was invented for this use case - the only complexity is that we want lots of sequences.

My vote is to still use sequences :) If someone wants to test sub-queries and locking with Andrew's code and see if it works, evidence may persuade me :)

pixeltrix:
One thing to note - with my code if we get to 100000 we get an error. Which is fairly important and not an impossible scenario :smile:

Anyway, here's a version that has three separate columns so we can lock the table to block any other reads while we're calculating the next index value:

``` ruby
require 'active_record'

ActiveRecord::Base.establish_connection(adapter: 'postgresql', encoding: 'unicode', database: 'defence_request')
ActiveRecord::Base.logger = Logger.new('/dev/null')

ActiveRecord::Schema.define do
  create_table :defence_requests, force: true  do |t|
    t.string  :first_name,  limit: 50
    t.string  :last_name,   limit: 50
    t.string  :dscc_prefix, limit: 4
    t.integer :dscc_index
    t.string  :dscc_suffix, limit: 1
    t.timestamps null: false
  end

  add_index :defence_requests, [:dscc_prefix, :dscc_index, :dscc_suffix], unique: true, name: 'index_defence_requests_on_dscc_number'
end

class DefenceRequest < ActiveRecord::Base
  DSCC_NUMBER_FORMAT = /\A[0-9]{9}[A-Z]\z/

  validates_numericality_of :dscc_index, less_than: 100_000

  class << self
    def find_by_dscc_number(number)
      unless number =~ DSCC_NUMBER_FORMAT
        raise ArgumentError, "Invalid DSCC number: #{number.inspect}"
      end

      find_by(dscc_prefix: number[0..3], dscc_index: number[4..8].to_i, dscc_suffix: number[9])
    end

    def find_by_dscc_number!(number)
      find_by_dscc_number(number) || raise_dscc_number_not_found(number)
    end

    private

    def raise_dscc_number_not_found(number)
      raise ActiveRecord::RecordNotFound, "Couldn't find request with DSCC number: #{number.inspect}"
    end
  end

  def dscc_number
    '%s%05d%s' % dscc_components if dscc_components.all?
  end

  def generate_dscc_number!
    unless persisted?
      raise RuntimeError, "Defence requests must be saved before generating a DSCC number"
    end

    dscc_number_generator.generate!
  end

  private

  def dscc_components
    [dscc_prefix, dscc_index, dscc_suffix]
  end

  def dscc_number_generator
    DSCCNumberGenerator.new(self)
  end
end

class DSCCNumberGenerator
  DSCC_SUFFIX = 'Z'

  def initialize(request)
    @request = request
  end

  def generate!
    with_lock do
      request.update!(
        dscc_prefix: prefix,
        dscc_index:  next_index,
        dscc_suffix: DSCC_SUFFIX
      )
    end
  end

  private

  def request
    @request
  end

  def with_lock
    transaction do
      lock_records
      yield
    end
  end

  def lock_records
    klass.lock(true).first
  end

  def current_index
    klass.where(dscc_prefix: prefix).maximum(:dscc_index) || 0
  end

  def next_index
    current_index + 1
  end

  def prefix
    @prefix ||= request.created_at.strftime('%y%m')
  end

  def transaction
    klass.transaction{ yield }
  end

  def klass
    @klass ||= request.class
  end
end

10.times do
  pid = Process.fork do
    ActiveRecord::Base.connection.reconnect!

    1000.times do
      request = DefenceRequest.create!(first_name: 'Andrew', last_name: 'White')
      request.generate_dscc_number!

      puts "Process: #{Process.pid}, DSCC No: #{request.dscc_number}\n"
    end
  end
end

Process.waitall
```

One thing to note is that this code is over twice as slow to generate 10,000 numbers as the sequence code. I also tried to lock just the rows that had the matching prefix but for some reason that didn't work - I don't think it would make that much difference to performance anyway since you're generally only going to be working on the current requests anyway.

PR #116, opened 2015-04-15
Converts application to use oAuth

pedrosmmoreira:
* Enable Postgres uuid-ossp extension
* Use Postgres uuid type
* Change Defence Request associations to use UUIDs
* Add Omniauth-dsds gem and due spec support


app/policies/defence_request_policy.rb
```diff
@@ -8,11 +8,11 @@ def initialize(user, scope)
     end
 
     def resolve
-      if user.cso?
+      if user.roles.include?("cso")
         scope.not_aborted
-      elsif user.cco?
+      elsif user.roles.include?("cco")
```
oskarpearson: Just wondering if there was a reason you didn't use the cco method from line 103 below? And similarly for cso etc.

app/views/defence_requests/index.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= flash_messages %>
 
-<h1><%= "#{t(current_user.role)}  #{t('dashboard')}" %></h1>
+<h1><%= "#{t(current_user.roles.first.to_s)}  #{t('dashboard')}" %></h1>
```
oskarpearson: Crying out for a helper method.

Also - would roles always have a guaranteed sort order? If it doesn't, the role displayed could vary from page to page.

pedrosmmoreira: Absolutely. I was planning on leaving that off until getting started on actually consuming profile info from the Auth app. This PR is a stop gap/setup of sorts, I guess.

spec/features/cco/dashboard_spec.rb
```diff
@@ -4,85 +4,62 @@
...
+      expect(page).to have_content "Acknowledged Req. Updated Solicitor Name"
     end
 
-    specify "can resend case details of an defence request to the assigned solicitor", js: true do
-      visit defence_requests_path
-
-      within ".accepted-defence-request" do
-        click_button "Resend details"
-      end
+    xspecify "can resend case details of an defence request to the assigned solicitor", js: true do
```
oskarpearson: I presume the xspecify here and below are intentional.

pedrosmmoreira: Yes, after discussion it was decided to have these specs pending, there are about 17 in total I believe.

db/migrate/20150413103214_add_solicitor_and_cco_uuids_to_defence_requests.rb
```diff
@@ -0,0 +1,6 @@
+class AddSolicitorAndCcoUuidsToDefenceRequests < ActiveRecord::Migration
+  def change
+    add_column :defence_requests, :solicitor_uuid, :uuid
+    add_column :defence_requests, :cco_uuid, :uuid
```
ansonK: Should be be indexing the *_uid columns?

ansonK:
Seems that the OAuth seeds for the service app don't match up with the seeds in the Auth app (application_id)
This prevents logging into the service app locally using the auth app as a provider

db/migrate/20150413103214_add_solicitor_and_cco_uuids_to_defence_requests.rb
pedrosmmoreira: Will we be internally retrieving requests by solicitor/cco uid? I didn't think we would, but I'll get the indexes in if we are.

robmckinnon:
I've just added a little commit to set oauth application id and secret to match those in auth app .env.development

pedrosmmoreira:
@ansonK: This should now be ready to merge as discussed.

novotnyjakub:
We just noticed that `db/seeds.rb` references `User` model, which has been removed by this PR. Could you please check what should stay in the seeds file and what should be removed?

pedrosmmoreira:
Thanks for the heads up @novotnyjakub, that should indeed go.

db/migrate/20150413103214_add_solicitor_and_cco_uuids_to_defence_requests.rb
robmckinnon: Shall we go with `_uid` as the prefix convention for UID fields, instead of `_uuid` ?

`_uid` feels better given we defined `uid` fields for Profile and Organisation in the [auth app schema](https://github.com/ministryofjustice/defence-request-service-auth/blob/master/db/schema.rb#L72).

robmckinnon: Opps - I didn't notice there is already another migration to rename them to `_uid`. So ignore above.

ansonK: > Will we be internally retrieving requests by solicitor/cco uid? I didn't think we would, but I'll get the indexes in if we are.

Yes we do need the *_uid columns indexed to handle the cases where we need to find all the Defence Requests assigned to a particular solicitor for example

chriscarter90:
@pedrosmmoreira I'm just going to push two little commits to this branch so we can get it merged:
* Adding indices to the UID fields on `DefenceRequest`
* Updating the gem version to `v0.6.1`

Then it should be merge-able - all good?

db/migrate/20150413103214_add_solicitor_and_cco_uuids_to_defence_requests.rb
chriscarter90: I've added indices for these two columns now.

chriscarter90:
I believe this PR is now ready to merge, however I am getting an odd test failure.

The test at `spec/features/cso/defence_request_management_spec.rb:19` seems to fail in the context of the suite or the file, but passes perfectly fine on its own. When I call `save_and_open_page` before the offending line, it fails with a completely blank `<head>` and `<body>`.

Any ideas @pedrosmmoreira?

chriscarter90:
Okay so if anyone is interested, it seems that downgrading from PhantomJS version `2.0.0` to `1.9.8` fixed that weird test failure.

chriscarter90:
Merging now. :shipit:

PR #118, opened 2015-04-20
wip: Organisation spyke spike

ansonK:
Turning this into a PR so I can comment on it


ansonK:
So looking this over there are 2 things that I don't like with this approach:

1. There is no clean way of setting the context of the request (in our case the AUTHORIZATION header) as spyke expects a class level connection object on Spyke::Base to be set.
We could do this in a before_action for each request I spose but then how to handle rails console sessions? If it took a block then at least it could be more dynamic ....

2. It dynamically pulls property names of objects from the response. There does not seem to be a way of defining a stable contact between each Spyke class and calling code. eg Trying to find an Organisation with an invalid uid:
```
class Organisation < Spyke::Base
end

organisation_uid = 'a24f1ded6cd84322bd43ed9199a4adb4'
user_token = "Bearer #{session['user_token']}"
Organisation.where(user_token: user_token).find(organisation_uid)
```
will return an Organisation with *no* properties, breaking the contract the Organisation model has with the rest of the application. As the properties are dynamically defined there is no way of having a stable contract with the consuming code (each Spyke model stores properties in an internal hash and controls access with a ```method_missing```, which basically calls ```attributes[name] if attributes.has_key?(name)``` which is useless if the API response was a 404

PR #119, opened 2015-04-21
Extracts Dashboard Resource

pedrosmmoreira:
This extracts a dashboard resource from DefenceRequestsController, enabling a simpler `DefenceRequestsController` and introducing a `Dashboard` object that can be extended without further overloading the views or the defence request model.

It also eliminates unnecessary queries that were made by `DefenceRequestsController#get_defence_request_scopes` and the corresponding duplicate calls to `render` in the view. The css classes passed down to render as `type` were also seemingly only used to facilitate testing. In a previous PR, it was ensured that each test creates only the set of data it needs and, therefore, these have become unnecessary as well.


app/controllers/base_controller.rb
```diff
@@ -3,9 +3,10 @@ class BaseController < ApplicationController
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
-  after_action :verify_authorized, except: :index
+  after_action :verify_authorized
```
ansonK: We should default to verifying that scopes are called for index actions, and other actions call authorize (or explicitly skip) so using Policies is enforced
eg
```
after_action :verify_authorized, :except => :index
after_action :verify_policy_scoped, :only => :index
```

pedrosmmoreira: Unless I'm mistaken, there are no index actions currently. I am happy to add the action back in, but to me it feels like adding confusion to an already not very clear authorization procedure. Should we reevalute this when the refactoring of 'DefenceRequestController' is complete?

ansonK: ah I had assumed (wrong) that the dashboards controller was the 'index' action as they display a list of records. Shouldn't the dashboards controller be index rather then show?

spec/support/dashboard_helper.rb
```diff
@@ -1,38 +1,5 @@
...
-    sleep(Settings.dsds.dashboard_refresh_milliseconds / 1000)
-    wait_for_ajax
-  end
-
-  private
-
-  def short_dashboard_refresh_milliseconds
-    SHORT_DASHBOARD_REFRESH_SECONDS * 1000
+  def refresh_dashboard
+    page.execute_script "window.jQuery.ajax({url: 'refresh_dashboard'})"
```
ansonK: this is bypassing the triggering of the dashboard refresh. The 'old' way being removed here is a better integration test as the dashboard refresh is being triggered by the same mechanism (just with a reduced wait time).

If we trigger the refresh explicitly like here we need a different test to test the actual triggering mechanism defined in app/assets/javascripts/dashboard_refresh_interval.js.erb

app/controllers/base_controller.rb
pedrosmmoreira: I think we are showing a dashboard. That happens to contain a list of records specific to a user, so in a sense we are also not exactly presenting a Defence Requests index. Does that make sense?

spec/support/dashboard_helper.rb
pedrosmmoreira: I had a lot of doubts about this one. I ended up going this way for three reasons:

1) We were always "conditioning" the triggering by changing interval settings or other strategies;
2) This will expose failures in the polling mechanism in a much cleaner manner and make the test deterministic - all other variations seem to intermittently fail.
3) The entire test to me seems incorrect. We are manually updating records in the db at an integration level. In a way that is much more concerning than triggering the refresh at will;

Should we discuss other options?

app/views/dashboards/show.html.erb
```diff
@@ -0,0 +1,20 @@
+<% content_for :javascript_includes do %>
+  <%= javascript_include_tag "dashboard_poller" %>
```
ansonK: If we change this to be
```
<%= javascript_include_tag "dashboard_poller" unless unless Settings.dsds.dashboard_refresh_disabled %>
```
then we can remove the erb tags from the dashboard_poller file so the javascript is consistent regardless of ENV variables

pedrosmmoreira: Good call, thanks!

ansonK: Ignore the above comment. I think the entire way the dashboard refresh is configured and triggered needs to be changed so is a separate task all to itself

PR #120, opened 2015-04-22
Extract transitions resource

pedrosmmoreira:
This extracts a resource responsible for managing the state transitions of defence requests.


ansonK:
Looks good.
We seem to be missing DefenceRequestTransition objects for the abort and complete transitions tho?

pedrosmmoreira:
Thanks! I'll be dealing with the transitions that require comments today, hopefully we can now make use of discrete strategies for those as well.

robmckinnon:
I noticed duplication in `AbortDefenceRequestsController` and `FinishDefenceRequestsController`, `abort_defence_request_form` and `finish_defence_request_form`, and the `DefenceRequestTransitions` classes.

@ansonK has explained to me that the workflows for abort and finish may diverge more next week, hence is comfortable with leaving duplication for now.

My suggestion is to place duplicated methods in mixin or base classes, that are included/extended elsewhere. Then bugs in duplicated methods could be fixed once in one place.

For example `set_cco()` has been implemented slightly differently in `DefenceRequestTransitions::Accept`
```ruby
def set_cco
    return true if defence_request.cco = user
end
```
 from `DefenceRequestTransitions::Acknowledge`
```ruby
def set_cco
    defence_request.cco_uid = user.uid
end
```
Should they be the same implementation? Is the first implementation correct?

app/models/defence_request_transitions/accept.rb
```diff
@@ -0,0 +1,31 @@
...
+    def complete
+      set_cco && transition_defence_request
+    end
+
+    private
+
+    attr_reader :defence_request, :requested_transition, :user
+
+    def set_cco
+      return true if defence_request.cco = user
```
robmckinnon: Should this be:   `defence_request.cco_uid = user.uid` ?

pedrosmmoreira:
Regarding the `set_cco` method, you are correct. They should be same and follow the guard form. I'll make that change.

Right now the strategies are quite similar. This was discussed offline and we are happy to keep them that way. There is, however, an important distinction to be made as per the interpretation of "duplication":

The strategies honor the same API contract, so we can have the advantage of polymorphism/duck typing. When you say you are worried about changing things in multiple places, or Shotgun Surgery, that is not correct, since you'd only have to implement changes to the affected strategy. There is an important distinction between duplicated text and duplicated knowledge, much as there is between DRY and finding a correct abstraction. Each method is the sole source of knowledge for an operation in each strategy. Thus, I do not believe we are looking at duplication. We might be looking at an interim state of finding the correct abstraction, but as Sandi Metz says, it is much preferable to work step by step in finding the correct abstraction, with the knowledge each requirement brings.

Using the Strategy pattern implies we considered the tradeoff between increasing overall complexity in the system in exchange for reducing it in individual components.
While a common suggestion, the use of Mixins is, essentially, a form of inheritance. The approach taken in this PR goes along with preferring composition over inheritance. It may be revisited once we get to a point of stability where we are comfortable with going with inheritance.

As a personal preference, I would rather not loose the flexibility granted by composition but will evidently revise with the team along the road.

app/models/defence_request_transitions/accept.rb
pedrosmmoreira: Yes, thanks for spotting that one :+1:

FriedSock:
@pedrosmmoreira I agree with what you are saying in principle, but do not agree with the exact functionality that each strategy is implementing. Reducing the flexibility may clear up this point of contention. 

Fore example I don't think that the `Acknowledge` strategy should have the dependency of the requested transition passed in to it, given that the transition is the thing we are switching strategies on. It should only be able to acknowledge requests. 

At this point there is basically only the aforementioned `set_cco` method that is shared across strategies with the same implementation. And I think we can agree that breaking out a mixin/superclass for a single method would be a rather premature refactor.

pedrosmmoreira:
@FriedSock I see your point. I think it would fine if each strategy knew the name of the method to call on the request, since at that point we've already switched. I opted for this design since we have a consistent data structure carried across, so we might has well keep it injected. Happy to revise that decision.

ansonK:
Here is an approach to the Acknowledge service class. A few changes:
* explicitly call acknowledge as there is no reason to make the transition name dynamic when the class is specifically for the Acknowledge transition (point raised by @FriedSock above)
* don't use exceptions for flow control. In this case can_acknowledge? removes the need to rescue
* I have kept the transition_params in the initializer instead of more explicit keyword args to allow calling to to pass in an arbitrary hash and the service object pick out the information it needs
```
module DefenceRequestTransitions
  class Acknowledge
    def initialize(transition_params)
      @defence_request = transition_params.fetch(:defence_request)
      @current_user = transition_params.fetch(:current_user)
    end

    def complete
      set_cco && acknowledge_defence_request 
    end

    private

    attr_reader :defence_request, :user

    def set_cco
      defence_request.cco_uid = current_user.uid
    end

    def acknowledge_defence_request
       run_transition && defence_request.save
    end

    def run_transition
      defence_request.can_acknowledge? && defence_request.acknowledge
    end
  end
end
```

pedrosmmoreira:
@ansonK I'm happy with that, if @FriedSock and @robmckinnon are also in agreement, I'll make those changes and push. Thank you all for your feedback on this!

ansonK:
What is missing from the Transitions service objects is error bubbling. When a transition fails (either due to a guard clause or a failed validation) then an error message must be bubbled up to the controller for display to the user.

Each of these service object could respond_to ```errors``` like below:

```
   if transition.complete
      redirect_to dashboard_path,
        notice: flash_message(requested_transition.to_sym, DefenceRequest)
    else
      redirect_to dashboard_path, error: transition.errors
    end
```
The ```errors``` response should be something that can be stored in the flash and rendered to the user

Thoughts?

pedrosmmoreira:
The ones with form objects make this easier to handle discreetly, I'll dig in to coming up with something to surface errors back up. My question is that, other than failing the transition, I am having some difficulty in seeing what other errors we'd pick up? Is it actually possible at each of these points to have objects in an invalid state which would be exposed by the call to `save!`?

ansonK:
It's completely possible for the defence request model passed into a service object to be invalid, or for validations inside the model to make the state invalid - the service object won't know until it calls ```save``` (same as in a controller really). 
eg Working off the example above there may well be something like this in the defence request:
```
validate :cco_uuid, presence: true, if: :acknowledged?
```

which would only take effect once ```defence_request.acknowledge``` had been called.

pedrosmmoreira:
Right, makes sense. I'll start investigating it. Thanks!

ansonK:
looks good, just needs a rebase

pedrosmmoreira:
Thanks! on it

FriedSock:
Could we remove the dependency on requested transition in the controllers as well as the transition models? I don't see the case for an Abort controller performing a different transition.

PR #124, opened 2015-04-24
Solicitors can see all Defence Requests for their Firm

robmckinnon:
* Implements: https://www.pivotaltracker.com/story/show/91086290
* Add `related_to_solicitor()` method to DefenceRequest that returns solicitor's firms defence requests.
* Assumes that solicitor's `organisation_uid` is always present and correct on defence requests that have `solicitor_uid == solicitor's uid`. This needs to be verified when we implement assigning a solicitor to a request.
* Having to write `solicitor.organisation_uids.first` to get solicitor `organisation_uid` - suggestions welcome for a better interface.
* Add `login_with()` helper to allow custom user fields to be mocked on profile.


app/models/defence_request.rb
```diff
@@ -10,8 +10,14 @@ class DefenceRequest < ActiveRecord::Base
   scope :not_draft, -> { where.not(state: "draft") }
   scope :ordered_by_created_at, -> { order(created_at: :asc) }
 
-  def self.has_solicitor(solicitor)
-    where(solicitor_uid: solicitor.uid)
+  class << self
```
pedrosmmoreira: What is the motivation for switching class method syntax here? This would, for instance, raise issues with Rubocop or other style checkers, adds noise to the class and seems a bit spurious?

robmckinnon: Think deleting has_solicitor() is a good idea. I'll check first that it is not used elsewhere.

spec/features/solicitor/dashboard_spec.rb
```diff
@@ -8,14 +8,16 @@
...
-      solicitor_uid: solicitor_user.uid
+      solicitor_uid: solicitor_user.uid,
+      organisation_uid: solicitor_user.organisation_uids.first,
+      offences: "Arson"
     )
-    not_accepted_defence_request = create :defence_request
+    not_accepted_defence_request = create :defence_request, offences: "Extortion"
 
-    login_with_role "solicitor", solicitor_user.uid
+    login_with_user solicitor_user
```
pedrosmmoreira: Any reason for changing this method name? I'm fine with passing less parameters around, but if feels the original name provided more context.

spec/models/defence_request_spec.rb
```diff
@@ -165,4 +165,27 @@
       end
     end
   end
+
+  describe "retrieve requests associated with solicitor's firm" do
```
pedrosmmoreira: It is preferable to maintain the 4 phase setup in tests as much as possible. Having scenarios that consist entirely of expectations obfuscates reading. In this particular case, there is not much to gain from extracting the before block, for instance.

pedrosmmoreira:
Some style related comments, otherwise looks good to me.

spec/features/solicitor/dashboard_spec.rb
robmckinnon: Method takes a user object as a parameter, so `login_with_user` seemed an appropriate name. The `login_with_role` helper method has not been deleted yet.

pedrosmmoreira: Ah, I see. How do you feel about refactoring `login_with_role` to accept the user object and keeping only one method? `login_with_user` feels slightly redundant while letting the test reader know that there are several types of user exposes important information, if that makes sense?

robmckinnon:
As agreed with @pedrosmmoreira and @ansonK :
* Renamed `login_with_user()` to `login_with()`
* Replaced `login_with_remote()` usage with `login_with()`

Should be ready for merge now.

spec/models/defence_request_spec.rb
robmckinnon: Have removed before block as suggested.

PR #127, opened 2015-04-27
[fix_failing_build_573] Fixed failing build

oskarpearson:
* Removed trailing whitespace
* Increased the size of Metrics/LineLength to match the actual codebase for now
* Replaced .kind_of?() -> .is_a?()

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -17,7 +17,7 @@ Metrics/AbcSize:
 # Offense count: 220
 # Configuration parameters: AllowURI, URISchemes.
 Metrics/LineLength:
-  Max: 135
+  Max: 138
 
 # Offense count: 1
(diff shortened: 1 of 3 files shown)


.rubocop_todo.yml
```diff
@@ -17,7 +17,7 @@ Metrics/AbcSize:
 # Offense count: 220
 # Configuration parameters: AllowURI, URISchemes.
 Metrics/LineLength:
-  Max: 135
+  Max: 138
```
novotnyjakub: Should we maybe disable this cop, rather than having it very long for backwards compatibility?

ansonK: We'll treat is as tech debt for the time being. Just get rubocop in and passing so nothing gets worse, then do spikes to fix certain things at 1 time (eg reduce all line lengths to something shorter)

PR #128, opened 2015-04-27
[fix_all_single_quotes] Double quote everything

oskarpearson:
Change all single quoted strings to double quotes, so that the codebase matches our coding standard.

Note that I've changed standard rails components in config/* and bin/*
etc. The reasoning for this is a little complicated:

1) I could exclude bin/* - but that way if someone creates a new bin/foo it won't be checked
2) I could exclude individual items like bin/bundle, bin/rails etc - but if someone does change it,
   their change won't be checked by rubocop

The same applies for other files, like things in config/* that are built by standard configs.

After some discussion with Anson, we've decided to replace all occurrences. The downside is that
if/when we do a rails upgrade, someone will have to compare a single-vs-double-quoted version of
the files in bin/* and similar to see if there are changes that need to be incorporated.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -100,12 +100,6 @@ Style/SpaceInsideBrackets:
 Style/SpaceInsideParens:
   Enabled: false
 
-# Offense count: 225
-# Cop supports --auto-correct.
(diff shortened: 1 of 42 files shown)


FriedSock:
:+1:

PR #130, opened 2015-04-27
Fix defence request table actions

FriedSock:
A bunch of actions were not being checked with
policies, now they are. 

Also merge actions back into the same partial 
because we are using policies to determine if you
should be able to see or perform an action, we
can keep this noise out of the view.


dp6ai:
LGTM

PR #132, opened 2015-04-28
Feature/solicitor dash tabs

dp6ai:
This cant be completed until https://www.pivotaltracker.com/story/show/90513334. 

Do not merge...


app/controllers/defence_requests_controller.rb
```diff
@@ -106,7 +106,7 @@ def queue
   private
 
   def associate_cco
```
pedrosmmoreira: We should be able to delete this method now that associating a cco occurs in the transition strategies. Should have spotted that, sorry.

spec/models/dashboard_spec.rb
```diff
@@ -1,4 +1,4 @@
-require "spec_helper"
+require "rails_helper"
```
pedrosmmoreira: Did using just "spec_helper" break? Nothing added in the test seems to cause that.

dp6ai: .blank?

pedrosmmoreira: ah, that skipped through the read. Is it worth the tradeoff? Probably it is only a matter of time until we have to switch it anyway.

app/controllers/dashboards_controller.rb
```diff
@@ -13,10 +18,28 @@ def refresh_dashboard
...
+  def closed_dashboard
+    @_closed_dashboard ||= dashboard(:finished, :aborted)
+  end
+
+  def set_dashboard_count
+    @active_count ||= active_dashboard.defence_requests.count
+    @closed_count ||= closed_dashboard.defence_requests.count
+  end
+
+  def dashboard(*additional_scopes)
```
pedrosmmoreira: Where do these additional scopes come from? Any reason why they shouldn't be loaded with the requests scoped by policy?

dp6ai: We need a way of scoping the dashboard DR's depending on when its being used. The policy scopes are the 'top level' scopes that a role can see. we need finer grain control than that.

app/models/dashboard.rb
```diff
@@ -4,18 +4,34 @@ def initialize(user, defence_requests_scoped_by_policy)
...
-    ordered_defence_requests
-  end
-
   def user_role
     user.roles.uniq.first
   end
 
+  def defence_requests(type)
+    type == :active ? active_defence_requests : closed_defence_requests
+  end
```
ansonK: We shouldn't need to do this here - the controller should initialise the dashboard model with the correct collection of Defence Requests

app/models/dashboard.rb
```diff
@@ -4,18 +4,34 @@ def initialize(user, defence_requests_scoped_by_policy)
...
 
   attr_reader :user, :defence_requests_scoped_by_policy
 
+  def active_defence_requests
+    if user_role == "solicitor"
+      scoped_defence_requests([:not_finished])
+    else
+      ordered_defence_requests
+    end
+  end
```
ansonK: This should be done in the Policy, not here. The policy has responsibility for enforcing constraints on which defence requests a user can view

pedrosmmoreira:
That was a pretty good pass moving responsibilities to the dashboard model :+1:

PR #133, opened 2015-04-29
Docker container repository 92999546

oskarpearson:
Until we have the secure repo and build system configured, we are going to use the public docker registry

* Reverted to the old way of handling multiple Dockerfiles, since CircleCI runs an older Docker release
* Improve caching of docker containers as per https://circleci.com/docs/docker#caching-docker-layers
* Build and push Docker containers on builds against certain branches
* Ensure that we set the Rails environment to Production before building assets
* Circleci overrides database.yml by default, which then gets committed to the Docker containers.
  Additionally it installs things in vendor and other places. We now force a git clean before building
  the container, so that only the things that should definitely be in the container are in the container.
* Add production database.yml config entry
* On building a docker container, force the filesystem timestamp to all be the same value. This helps
  ensure cacheability, since docker considers a timestamp difference to mean a change, and
  git doesn't preserve timestamps across builds.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,19 @@
 machine:
   ruby:
     version: 2.2.2
+  services:
+    - docker
 test:
   override:
     - bundle exec rake --trace
+deployment:
+  create_containers:
+    branch: [master, docker_container_repository_92999546]
+    commands:
+      # Log into the docker container. Note that this requires these environment variables
+      # to be stored in the CircleCI environment config
+      - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
+      # Ensure that the docker container built contains absolutely nothing unexpected
+      # that might have been installed by CircleCI (eg an updated database.yml)
+      - (git clean -fd ; git reset --hard) || true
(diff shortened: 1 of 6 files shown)


circle.yml
```diff
@@ -1,6 +1,19 @@
...
+deployment:
+  create_containers:
+    branch: master
+    commands:
+      # Log into the docker container. Note that this requires these environment variables
+      # to be stored in the CircleCI environment config
+      - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
+      # Ensure that the docker container built contains absolutely nothing unexpected
+      # that might have been installed by CircleCI (eg an updated database.yml)
+      - (git clean -fd ; git reset --hard) || true
```
FriedSock: If the purpose of the `|| true` is to force the command exit code to be 0, I think we could just use the `-q` flag for `git reset`?

FriedSock:
LGTM, just a couple of questions

circle.yml
oskarpearson: It's probably best for us to actually generate an error if git clean or reset can't do their job. Let me fix that. Thanks.

PR #134, opened 2015-04-29
Solicitor dashboard list items

novotnyjakub:
This adds a responsive list of defence request to the solicitor dashboard. 

When accessed on a mobile device, it looks like this: 
screenshot_mobile

Because we don't have design for desktop just yet, for the moment the boxes are displayed 3 per row on larger screens:
screenshot_desktop


novotnyjakub:
Ok, I need to fix one more thing - the dashboard refresh.

novotnyjakub:
Ok, now it should be good to go!

PR #135, opened 2015-04-30
Refactor defence request policy

dp6ai:


app/policies/policy_context.rb
```diff
@@ -0,0 +1,12 @@
+class PolicyContext
+  attr_reader :record
+
+  def initialize(record, user)
+    @record = record
+    @user = user
+  end
+
+  def policy_class
+    "#{@user.roles.uniq.first}DefenceRequestPolicy".classify
```
oskarpearson: Is roles guaranteed to be ordered? Otherwise this could change from request to request.

Also: should we not have a method on user that will return this in a consistent way? I'm concerned we know too much about the implementation of user, and will have this construct all over our codebase.

app/views/dashboards/show.html.erb
```diff
@@ -6,11 +6,11 @@
 
 <h1><%= "#{t(@dashboard.user_role)}  #{t('dashboard')}" %></h1>
 
-<% if policy(:defence_request).create? %>
+<%# if policy(:defence_request).create? %>
```
oskarpearson: Is this supposed to be commented? If not, should we not just remove the if statement?

app/views/defence_requests/form_partials/_solicitor.html.erb
```diff
@@ -2,6 +2,6 @@
   <h3><%= t('solicitor_details') %></h3>
   <div class="details">
     <%= render('defence_requests/form_partials/solicitor_details', f: f) %>
-    <%= render('defence_requests/form_partials/solicitor_time_of_arrival', f: f) if @policy.solicitor_time_of_arrival? %>
+    <%= render('defence_requests/form_partials/solicitor_time_of_arrival', f: f) %> #if @policy.solicitor_time_of_arrival? %>
```
oskarpearson: Should this actually be commented?

app/policies/policy_context.rb
dp6ai: I agree this needs looking at in more detail but probably not here. Ill stick a story in the icebox

app/views/dashboards/show.html.erb
dp6ai: good spot

app/views/defence_requests/form_partials/_solicitor.html.erb
dp6ai: again.

app/policies/policy_context.rb
novotnyjakub: Wouldn't it be better to have more explicit way of constructing these? That you'd have to request which policy do you want and for which user. Ie:

```ruby
def initialize(record, policy_type, user)
```

I think we have already identified that we'll have multiple roles for a user.

app/controllers/defence_requests_controller.rb
```diff
@@ -1,14 +1,17 @@
...
   before_action :find_defence_request, except: [:new, :create]
-  before_action :set_policy
+  before_action :set_policy_with_context, only: [:show, :new, :create, :edit, :update, :resend_details, :solicitor_time_of_arrival]
   before_action :new_defence_request_form, only: [:show, :new, :create, :edit, :update, :solicitor_time_of_arrival]
 
-  before_action ->(c) { authorize defence_request, "#{c.action_name}?" }
+  before_action ->(c) { authorize PolicyContext.new(defence_request, current_user), "#{c.action_name}?" }
 
   def show
+    set_policy_with_context
```
novotnyjakub: These are being set in the `before_action` filter already.

app/policies/policy_context.rb
dp6ai: I think this is probably something we will need in the future and would be a relatively easy change now but im not sure its a change we should make now.

app/controllers/defence_requests_controller.rb
dp6ai: :+1:

app/policies/policy_context.rb
novotnyjakub: I don't have much experience with Pundit, so I'm going through the code trying to understand it. But if get it correctly, this approach won't allow a user to have 2 different roles on the same controller, will it? Cause the `before_action` filter checks a single policy against the requested action. So if I'm a _solicitor_ and _company admin_, one of the roles does allow access to an action and the other doesn't, it won't work as expected.

dp6ai: @novotnyjakub as discussed offline: The multiple role issue can probably be handled in the short term by finding the role with the 'highest' level of permissions, any other roles will then be a subset of those roles. With regards to the repetition of action names across controllers, agreed that most of the small controllers can be given a single policy clause 'close_dr' for instance which would cover the whole controller rather than individual actions and we assume that the standard CRUD actions relate to the defence_request_controller as the policies are named by the DR model, ie CcoDefenceRequestPolicy which is a side effect of the pundit implementation

ansonK: This is ok for the time being.
It will need to change in the near future once we decide how to handle the "super" or "admin" version of each cso, cco and solicitor role

ansonK:
lgtm

PR #136, opened 2015-05-01
WIP: Refactor features to use page object

novotnyjakub:
This PR is not to be merged, but as a discussion about this approach to feature testing. 

There's a couple of reasons for using similar approach:

1. With `page.have_content`, there's a big risk of false positives, because of ambiguity. For example testing that a time 11:50 is on the page, it can be anywhere on the page.

2. The HTML is abstracted from the tests, which makes them shorted and more readable. Also if the markup changes, only the page objects are changed, not feature files.

3. Sections can be re-used or replaced. So for example when testing different version for mobile or desktop (could be different markup), only the page object definition will be swapped, features can be the same.

4. It helps focusing on relevant page content and avoiding negative testing (something is not on the page).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,7 @@ group :test do
   gem "poltergeist"
   gem "shoulda-matchers", require: false
   gem "simplecov", require: false
+  gem "site_prism", require: false
   gem "webmock", require: false
 end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -292,6 +292,9 @@ GEM
       multi_json (~> 1.0)
       simplecov-html (~> 0.9.0)
     simplecov-html (0.9.0)
+    site_prism (2.7)
+      addressable (>= 2.3.3, < 3.0)
+      capybara (>= 2.1, < 3.0)
     slop (3.6.0)
     sprockets (3.0.2)
       rack (~> 1.0)
@@ -371,6 +374,7 @@ DEPENDENCIES
   sdoc (~> 0.4.0)
   shoulda-matchers
   simplecov
+  site_prism
   sucker_punch (~> 1.0)
   terminal-notifier-guard
   transitions
--- a/spec/features/solicitor/dashboard_spec.rb
+++ b/spec/features/solicitor/dashboard_spec.rb
@@ -3,43 +3,42 @@ require "rails_helper"
 RSpec.feature "Solicitors viewing their dashboard" do
   include DashboardHelper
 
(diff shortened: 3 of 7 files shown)


pixeltrix:
Isn't this still really using CSS selectors to specify where the content is? Other than masking the fact that we're using CSS selectors what else does this give us. I can see the benefit of giving names to the selectors but that could be just as simple as setting some constants in the spec.

I'm not sure whether the UI of the app is complex enough to warrant this level of abstraction but I'll leave it up to @ansonK to make the final decision.

novotnyjakub:
Sorry, I pressed the wrong button..

novotnyjakub:
The page object itself is mostly for not polluting the feature specs and keeping them simple and semantically focused on what they should test. Also reusability, because each page will be under test in multiple features (from the top of my head at least for each individual user role). Defining constants in each feature test again would be a maintenance overhead. 
For example I had to change 3 feature files to change a text of a button, even though I haven't actually changed any features. This is the kind of help the page object provides. 
Next benefit is that we'll (most probably) have a responsive and desktop version of the same view in some cases. Simply because sometimes table is the most appropriate desktop view, but doesn't fit at all on mobile. For this we'll be able to use the same feature, just switching the html abstraction to a different markup. I agree that this is way in the future.

pedrosmmoreira:
You raised a good point regarding the text. If our UI relies on I18n, so should our tests for instance. I like the page objects approach and used it myself but I do not think we are at a point that it is warranted.

The UI is in a heavy flux of change and we feel the pain of that change in the specs - that can tell you that we need abstraction to protect us from that change but it also highly indicates that, at this point in time, you are almost certainly choosing the wrong abstraction.

Also, I am not sure I understand how you are defining semantic value: for instance, I heavily rely on reading the setup for each scenario and find that to be of highly valuable. This change moves a lot of setup into `let`s, which can very easily make tests less deterministic. In many cases, relying on a `let!` and expecting that data is under your control is dangerous - it assumes you control bringing that object into existence at a very particular point. All it takes is another dependent caller at another point for you to have that object present much earlier than you tought.

I did a cleanup of the entire test suite so we could stop relying on scoping expectations to selectors, for instance, and managed to stop several intermittent failures. If we agree as a team to write page objects, I won't object in any way, but I wanted to leave these points for consideration in making the decision.

novotnyjakub:
You're right it's better for feature tests to be more explicit about setup and the individual steps. I'll think about it and push up a change when I finish what I'm doing now.

PR #138, opened 2015-05-05
Fix generating dscc_number on acknowledge

FriedSock:
This somehow got lost in a rebase.

### DIFF
--- a/app/models/defence_request_transitions/acknowledge.rb
+++ b/app/models/defence_request_transitions/acknowledge.rb
@@ -18,7 +18,7 @@ module DefenceRequestTransitions
     end
 
     def acknowledge_defence_request
-      run_transition && defence_request.save!
+      run_transition && defence_request.save! && defence_request.generate_dscc_number!
     end
 
     def run_transition
--- a/app/models/defence_request_transitions/acknowledge.rb
+++ b/app/models/defence_request_transitions/acknowledge.rb
@@ -17,6 +17,9 @@ module DefenceRequestTransitions
       return true if defence_request.cco_uid = user.uid
     end
 
+    # Note: In the unlikely event that we cannot generate a dscc_number the dr
+    # is going to end up in a broken state. We will move this into a delayed
+    # job where we will handle this case.
     def acknowledge_defence_request
       run_transition && defence_request.save! && defence_request.generate_dscc_number!
     end
--- a/spec/features/cco/defence_request_management_spec.rb
+++ b/spec/features/cco/defence_request_management_spec.rb
@@ -36,12 +36,15 @@ RSpec.feature "Call Center Operatives managing defence requests" do
 
       specify "can acknowledge the request" do
         cco_user = create :cco_user
-        create :defence_request, :queued, cco_uid: cco_user.uid
+        dr =  create :defence_request, :queued, cco_uid: cco_user.uid
 
         login_with cco_user
         click_button "Acknowledge"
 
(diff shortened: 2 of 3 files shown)


app/models/defence_request_transitions/acknowledge.rb
```diff
@@ -18,7 +18,7 @@ def set_cco
     end
 
     def acknowledge_defence_request
-      run_transition && defence_request.save!
+      run_transition && defence_request.save! && defence_request.generate_dscc_number!
```
pedrosmmoreira: How do you feel about extracting a method along the lines of `save_and_generate_dscc_number`? Even though the name hints at trouble, it feels like the defence request should handle this. What do you think?

spec/features/cco/defence_request_management_spec.rb
```diff
@@ -36,12 +36,15 @@
...
-        create :defence_request, :queued, cco_uid: cco_user.uid
+        dr =  create :defence_request, :queued, cco_uid: cco_user.uid
 
         login_with cco_user
         click_button "Acknowledge"
 
         expect(page).to have_content "Defence Request successfully acknowledged"
+
+        click_link "Show"
+        expect(page).to have_content "#{dr.dscc_number}"
```
novotnyjakub: I'm wondering if we should also test that the transition happened? Ie. that the `dr` is in acknowledged state?

pedrosmmoreira: That would not be visible in the UI, right? Other than the number being there.

app/models/defence_request_transitions/acknowledge.rb
FriedSock: Good suggestion, I think we want to disuade people from calling `defence_request.generate_dscc_number!` without save as much as possible, this would allow us to make it private.

pedrosmmoreira: :+1:

spec/features/cco/defence_request_management_spec.rb
FriedSock: I guess we are trusting the UI to not be lying to us. I would rather keep assertions about models themselves out of the acceptance tests if possible.

novotnyjakub: Cool, if we do that though, we should add this to the transition model test. It got missed in the first place, which means we have a gap somewhere.

pedrosmmoreira: Or better yet, to the `defence_request` model, since we are extracting it there :)

novotnyjakub: It's tricky. In my opinion we should be testing it where we do the operations. The `defence_request` model already has the transition condition. But because we test the individual transitions in isolation, I think it should be tested there.

pedrosmmoreira: I see your point. Although we wouldn't be doing more that spying on a message send. In that regard, the transition does not _perform_ the work or exercise that functionality - nor should it. It feels the proper place for the test would be in the unit that actually does the work. What do you think?

FriedSock: I think I agree with @novotnyjakub, and have also changed my opinion of where the code belongs. Will update transition specs.

pedrosmmoreira: Cool, no worries. LGTM anyway.

PR #142, opened 2015-05-05
Finish -> Complete, no more mandatory comment

FriedSock:
Rename the "Finish" action to "Complete". We also no longer need a comment to be supplied on completion, so remove that functionality.


chriscarter90:
:shipit:!

PR #143, opened 2015-05-05
Auth failure route

ansonK:
Updates omniauth-dsds gem to the version that checks a users roles when logging in

Implements a public failure route thats users see when accessing the rota when they have no roles in the rota app.
Uses the omniauth default for failure route ```/auth/failure```

Depends on omniauth-dsds gem checking that current_user has roles for the application as a dependency: https://github.com/ministryofjustice/defence-request-service-omniauth-dsds/compare/v0.6.2...master


chriscarter90:
Cool :+1: :shipit:

PR #144, opened 2015-05-06
Feature/solicitor dashboard tabs

dp6ai:


app/helpers/dashboard_helper.rb
```diff
@@ -0,0 +1,6 @@
+module DashboardHelper
+
+  def active_tab?(action)
+    "active_tab" if params[:action] == action
+  end
+end
```
FriedSock: I don't like the idea of a method ending with `?` that is not a predicate. Could we not have a `tab_class` method that returns all of the classes for the action or something?

dp6ai: good point

PR #146, opened 2015-05-07
Add organisation uid to defence request

novotnyjakub:
The purpose of this PR is to temporarily assign a defence request to the first available law firm. The functionality in the `Acknowledge` transition will be replaced later using the Rota service. 

Few notes on this PR, discussed and agreed with Anson: 
* I refactored the `Acknowledge` transition test to use the actual `DefenceRequest`, instead of stubs. The reason is that we don't have any other test covering this core domain logic. That should make sure we don't miss anything for that transition. I'd later refactor the other transitions to do the same. 
* In specs, we need to be able to mock more complex behaviour of the Auth Api client. So I made a configurable client, when from each spec, a response can be stubbed for each set of filter parameters. 

Let me know your thoughts!


app/controllers/transition_defence_requests_controller.rb
```diff
@@ -34,6 +34,7 @@ def transition_params
       defence_request: defence_request,
       transition_to: requested_transition,
       user: current_user,
+      auth_token: session[:user_token]
```
pedrosmmoreira: How do you feel about passing the instantiated api client down from the controller? Injecting the dependency that way can stop us referencing a constant all the way down into the transition.

lib/service_registry.rb
```diff
@@ -0,0 +1,17 @@
+class ServiceRegistry
```
pedrosmmoreira: The service registry implementation differs slightly from the one in Rota, do you think it would be a good idea to keep them the same?

spec/models/defence_request_transitions/acknowledge_spec.rb
```diff
@@ -1,51 +1,71 @@
...
-require "transitions"
+require "rails_helper"
 
-RSpec.describe DefenceRequestTransitions::Acknowledge, "#complete" do
-  it "transitions the defence request to acknowledge state" do
-    defence_request = spy(:defence_request)
-    user = spy(:user, uid: SecureRandom.uuid)
+RSpec.describe DefenceRequestTransitions::Acknowledge, "#complete", :mock_auth_api do
+  let(:defence_request) { create(:defence_request, :duty_solicitor, :queued, dscc_number: nil ) }
+  let(:user) { spy(:user, uid: SecureRandom.uuid) }
```
pedrosmmoreira: Any reason for making user a spy? We are not running assertions against the object and, since we are depending on Rails, it seems we might as well use a full fledged object here.

spec/support/mock_auth_api_helper.rb
```diff
@@ -0,0 +1,54 @@
+module MockAuthApiHelper
+  MockAuthApiNotSetup = Class.new(Exception)
+
+  def mock_auth_api_client(setup)
+    @old_service = ServiceRegistry.service(:auth_api_client)
```
pedrosmmoreira: We shouldn't need to keep a reference to the registered service as we should be able to just assign the mock to the registered key. Teardown is implicit as it gets set to the actual client in the initializer.

app/controllers/transition_defence_requests_controller.rb
novotnyjakub: This is temporary as I'd like to put the `auth_token` on to the `current_user` object. The current user is created from the token anyway and we won't have to pass it around separately.

I thought about passing the client from the controller, but then none of the other transitions need it. So it seems unnecessary to instantiate it every time.

lib/service_registry.rb
novotnyjakub: Yeah, this is a result of me trying out different things. I agree it's a good idea, but I'm not sure how to achieve it. Cause it would be C&P, so it'll get out of sync anyway.

spec/models/defence_request_transitions/acknowledge_spec.rb
novotnyjakub: Good spot, I didn't touch that at all. I'll use the factory for this.

app/controllers/transition_defence_requests_controller.rb
pedrosmmoreira: Good point. I do think `auth_token` lives separately so we can ensure it is fresh, at least from looking at the implementation of the check for current user. We could slightly change the strategy builder to handle this, but if we have no additional information at this point, we might as well keep the way you've implemented it.

lib/service_registry.rb
pedrosmmoreira: Ok.

app/controllers/transition_defence_requests_controller.rb
novotnyjakub: The `current_user` is build from the `auth_token` on every request. And `current_user` is read only, so there's no reason we couldn't put it there. At least we'll have all user related information available in one place. But we definitely can do it later.

pedrosmmoreira: Ah, I might have misunderstood the intent of the design then.

spec/support/mock_auth_api_helper.rb
novotnyjakub: So I've verified that the initialisers are not run before every test. I'll leave this manual tear down here for the moment.

novotnyjakub:
Cool, changes done, if you're happy, please merge ;-)

pedrosmmoreira:
LGTM

PR #148, opened 2015-05-07
Allow solicitor to see organisation defence request detail

novotnyjakub:
This may as well change in the future, when we have the roles defined more clearly. But for the moment solicitor has to be able to see the detail.


dp6ai:
LGTM

novotnyjakub:
Cool, do you think you can merge please?

PR #149, opened 2015-05-08
Refactor common defence request controller code into mixin

robmckinnon:
As suggested by @dp6ai and discussed with @FriedSock refactoring to make more use of defence request module mixin across controllers.

Up for discussion.

### DIFF
--- a/app/controllers/concerns/defence_request_authorization.rb
+++ b/app/controllers/concerns/defence_request_authorization.rb
@@ -1,11 +1,10 @@
 module DefenceRequestAuthorization
 
-  def find_defence_request
-    @defence_request = DefenceRequest.find(defence_request_id)
-  end
+  private
 
-  def set_policy_with_context
-    @policy ||= policy(PolicyContext.new(defence_request, current_user))
+  def find_defence_request
+    defence_request_id = params[:defence_request_id] || params[:id]
+    @defence_request = DefenceRequest.find(defence_request_id)
   end
 
   def defence_request
@@ -13,7 +12,13 @@ module DefenceRequestAuthorization
   end
 
   def new_defence_request_form
-    @defence_request_form ||= DefenceRequestForm.new(@defence_request)
+    @defence_request_form ||= DefenceRequestForm.new(defence_request)
+  end
+
+  def authorize_defence_request_access action
+    policy_context = PolicyContext.new(defence_request, current_user)
+    @policy ||= policy(policy_context)
+    authorize policy_context, "#{action}?"
   end
 
 end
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/defence_request_authorization.rb
```diff
@@ -1,19 +1,24 @@
 module DefenceRequestAuthorization
 
+  private
+
   def find_defence_request
+    defence_request_id = params[:defence_request_id] || params[:id]
```
novotnyjakub: I'm not sure about this, seems it could have unpredictable behaviour? `:id` is pretty generic. Maybe the individual controllers should decide where the `id` comes from?

app/controllers/concerns/defence_request_authorization.rb
```diff
@@ -1,19 +1,24 @@
...
 
-  def set_policy_with_context
-    @policy ||= policy(PolicyContext.new(defence_request, current_user))
-  end
-
   def defence_request
     @defence_request ||= DefenceRequest.new
   end
 
   def new_defence_request_form
```
novotnyjakub: This doesn't have anything to do with authorisation, should it be somewhere else?

app/controllers/concerns/defence_request_authorization.rb
robmckinnon: Seems sensible to let controllers decided on `id`. I experimented with a method in each controller to do that. Could go back to that approach.

FriedSock:
Agree with the comments from @novotnyjakub, otherwise lgtm

app/controllers/concerns/defence_request_authorization.rb
robmckinnon: Shall we leave this method here for now? Might be overkill to create a new module for a single method. Other option is to repeat it across controllers.

FriedSock: Id prefer to have it duplicated in the controllers than have it grouped in this module with the other auth stuff, that would be the least surprising thing.

novotnyjakub: In that case I'd personally prefer to have it in the `BaseController` rather than here. Because this is an authorisation concern only. 
But I'll leave this up to you, maybe ask another opinion. I'm not insisting on this to be done now :)

robmckinnon: Shall I rename the module to `DefenceRequestHelpers`?

robmckinnon:
Ok, made changes as suggested. Merging.

PR #154, opened 2015-05-13
[moj_ci_and_registry] Added support for an internal MOJ CI and Registry

oskarpearson:
* support for an environment-supplied DOCKER_PUBLISH_PREFIX parameter to Makefile
* Make minor changes to the Docker file to support concurrent builds on the same machine
  at the same time
* Add a test environment, so that we can run tests in docker containers


ansonK:
looks good to me

PR #155, opened 2015-05-13
Cleanup defence request

novotnyjakub:
This work removes fields from `DefenceRequest`, which we know won't be used even beyond MVP. Solicitor and custody suite will be referenced via uuid. 

I also had to make a few changes to the ui, to accommodate this. Quite a few feature tests tested results by checking if a solicitor's name is on the page. I replaced these with detainee's name, but even though I'm not sure it's the most reliable way. But that's not a purpose of this PR. 

I've tested the existing flow in the app using different roles and it works. There are some bugs and obviously a lot will change in the next weeks, but the current functionality should be preserved.


ansonK:
looks ok to me

robmckinnon:
Looks good, merging.

PR #156, opened 2015-05-13
Open and completed tabs

FriedSock:


novotnyjakub:
I'm thinking, if we have the same behaviour for different users, should we use `shared_examples` for feature tests to be clear that the behaviour is shared?

FriedSock:
@novotnyjakub I am not against that, but I know there are different opinions about how we should structure our examples. Eg some people like using lets and before blocks, others like to set up the world entirely within each example. I usually try to keep consistency with the other examples in the file

app/controllers/dashboards_controller.rb
```diff
@@ -27,15 +26,8 @@ def dashboard
   end
 
   def set_defence_requests
-    if @dashboard.user_role == "solicitor"
-      @active_defence_requests = @dashboard.defence_requests.active
-    else
-      @active_defence_requests = @dashboard.defence_requests
-    end
+    @active_defence_requests = @dashboard.defence_requests.active
     @closed_defence_requests = @dashboard.defence_requests.finished
```
novotnyjakub: We always load both sets of defence requests, even though only one is used. Same for the counts bellow.

app/controllers/dashboards_controller.rb
```diff
@@ -3,7 +3,6 @@ class DashboardsController < BaseController
   before_action :set_policy_with_context, except: [:refresh_dashboard]
   before_action :set_dashboard
   before_action :set_defence_requests, only: [:show, :closed]
-  before_action :set_dashboard_count, only: [:show, :closed]
```
novotnyjakub: I'm thinking that maybe dashboard could be `resources` instead of `resource` and that we could have different types like that. It would be more RESTful and I think maybe cleaner. It could help also with the pre-loading of different lists of defence requests, because it could be a scope with the same name (or something like that).

But that's just an idea, I'm not insisting.

pedrosmmoreira:
Have

app/controllers/dashboards_controller.rb
pedrosmmoreira: I don't quite get that, can you provide an example?

app/controllers/dashboards_controller.rb
FriedSock: yeah good point, will refactor that.

app/controllers/dashboards_controller.rb
novotnyjakub: Sure. So in `routes.rb`: 

```ruby
  resources :dashboard, only: :show
```

Then in the `dashboards_controller.rb`:

```ruby
  def show
    type = params[:id]
    set_defence_requests(type)
    set_dashboard_count(type)
  end
```

These methods then would check that the type is allowed and use the appropriate scopes on `DefenceRequest`. 

I hope that makes sense :)

pedrosmmoreira: Ah, I see. So, making it a singular resource seems unrelated to that - there is nothing that stops that happening or being Restful. You'd be showing a dashboard. Regarding the scoping: we should probably take all the necessary measures to only call scopes on `DefenceRequest` within a policy. Anything after that should just traverse the set returned. 

I like the notion of passing an `id` in order to have the open/closed dashboard on a different url, for instance, though.

novotnyjakub: Yeah, I'm sure that the policy check could work with the `id` automatically. I'm not too familiar with the policy implementation, but the `id` could be considered a scope and the policy context would automatically verify that the user (and type) is authorised.

FriedSock: I agree with this idea. The show action should not be reserved for showing a particular type of dashboard.

pedrosmmoreira: Right, that is not the issue I was referring to. We should have one single point that calls `policy_scope` with the user. From then on, we should only work with that whitelisted set of records - since they should always include every record we need. Does that make sense?

app/controllers/dashboards_controller.rb
```diff
@@ -27,15 +26,8 @@ def dashboard
...
 
   def set_defence_requests
-    if @dashboard.user_role == "solicitor"
-      @active_defence_requests = @dashboard.defence_requests.active
-    else
-      @active_defence_requests = @dashboard.defence_requests
-    end
+    @active_defence_requests = @dashboard.defence_requests.active
     @closed_defence_requests = @dashboard.defence_requests.finished
-  end
```
ansonK: Why are we setting many instance variables on the controller when we have a form object?

I would prefer only setting 1 instance variable in the controller (the form object), with everything the view needs being supplied by the form object

ansonK:
Looks like the css for the tabs is not consistent - switching between active and closed tabs makes the tabs appear different (loses border)

ansonK:
Url's should be ```/dashboard/active``` and ```/dashboard/closed``` as each user can only ever view 1 type of dashboard so it makes more sense to have it as a singular resource

config/routes.rb
```diff
@@ -2,8 +2,7 @@
   mount JasmineRails::Engine => "/specs" if defined?(JasmineRails)
   root "dashboards#show"
 
-  resource :dashboard, only: [:show]
-  get "/dashboards/closed", to: "dashboards#closed", as: "dashboard_closed"
+  get "/dashboards/show(/:id)", to: "dashboards#show", as: :dashboard
```
ansonK: Url's should be ```/dashboard/active``` and ```/dashboard/closed``` as each user can only ever view 1 type of dashboard so it makes more sense to have it as a singular resource

Fine to keep the root url for each user as an alias for ```/dashboard/active```

ansonK:
:+1:

PR #157, opened 2015-05-15
[RFC] Reload resistant registry

FriedSock:
In development mode, rails likes to reload
classes. Unfortunately this means our registry
gets clobbered, and a new singleton object is
instantiated - losing any services we registered
in initialisers. 

Keeping a separate class to use as the store
protects the service hash itself from being 
recreated (due to rails only expecting the
ServiceRegistry class to be residing in that
file)


FriedSock:
using a simpler method in https://github.com/ministryofjustice/defence-request-service/pull/160

PR #161, opened 2015-05-18
Add logout link

dp6ai:
#dependent on PR https://github.com/ministryofjustice/defence-request-service-auth/pull/77

https://www.pivotaltracker.com/story/show/93266326


db/schema.rb
```diff
@@ -85,4 +85,15 @@
...
+    t.string   "name"
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
+
+  create_table "roles_users", force: :cascade do |t|
+    t.integer "role_id"
+    t.integer "user_id"
+  end
+
```
ansonK: Why do we need roles and roles_users in the service app?

PR #163, opened 2015-05-20
CSO can mark name, address and date of birth as 'not given'

dp6ai:
WIP for handover to @FriedSock.

DO NOT MERGE


novotnyjakub:
I think this is now ready to review / merge. 

Functionally it's as the story defines, but visually it needs a bit more treatment. After a discussion with Kate, I think we should slightly improve the whole form in terms of font size and paddings. I'll do it in a separate PR / story.

Another thing which was agreed is to refactor the form javascript to not be the procedural code in the partials, but to be more encapsulated 'object' like behaviour, which will be 1) tested using Jasmine 2) initialised in the views where needed.

app/forms/defence_request_form.rb
```diff
@@ -4,17 +4,21 @@ class DefenceRequestForm
...
-                          :detainee_age, :offences, :comments, :appropriate_adult,
-                          :appropriate_adult_reason, :house_name, :address_1, :address_2, :city, :county, :postcode,
-                          :circumstances_of_arrest, :investigating_officer_name,
+  DELEGATED_ATTRIBUTES =  :detainee_name, :gender, :offences, :comments, :appropriate_adult, :appropriate_adult_reason,
+                          :detainee_address, :circumstances_of_arrest, :investigating_officer_name,
                           :investigating_officer_shoulder_number, :investigating_officer_contact_number,
                           :fit_for_interview, :unfit_for_interview_reason, :interpreter_required, :interpreter_type
 
+  VIRTUAL_ATTRIBUTES =  :detainee_name_not_given, :detainee_address_not_given, :date_of_birth_not_given
+
```
dp6ai: if these are now persisted attributes i think you can drop the VIRTUAL_ATTRIBUTES usage?

app/forms/defence_request_form.rb
```diff
@@ -4,17 +4,21 @@ class DefenceRequestForm
...
   DELEGATED_ATTRIBUTES.each do |attr_name|
     def_delegator :@defence_request, "#{attr_name}_before_type_cast", attr_name
   end
 
+  VIRTUAL_ATTRIBUTES.each do |attr_name|
+    def_delegator :@defence_request, attr_name, attr_name
+  end
+
```
dp6ai: And this...

novotnyjakub:
I've also updated the copy as per discussion with Kate.

robmckinnon:
Looks good to merge. Agreed with @novotnyjakub to work with him on javascript refactoring in a separate PR.

FriedSock:
Nitpick, but could we remove the comments in `_not_give_js.html.erb`? they aren't really adding anything. Maybe that could be part of the JS refactor.

novotnyjakub:
@FriedSock all the partials will be removed and the JS for the form will be moved to a class (of some sort) and tested. That's the refactoring @robmckinnon mentioned. We're starting on that right away, I just wanted this PR to be merged as from the feature perspective it's complete.

PR #165, opened 2015-05-28
Style cso dashboard

FriedSock:
Stories: https://www.pivotaltracker.com/story/show/93690552
and https://www.pivotaltracker.com/story/show/93690566


spec/models/dashboard_spec.rb
```diff
@@ -1,3 +1,4 @@
+require "rails_helper"
 require_relative "../../app/models/dashboard"
```
pedrosmmoreira: we can lose the relative here, now that we are using rails_helper

app/models/dashboard.rb
```diff
@@ -40,7 +41,20 @@ def closed_defence_requests
   attr_reader :user, :defence_requests_scoped_by_policy
 
   def ordered_defence_requests
-    defence_requests_scoped_by_policy.ordered_by_created_at
+    defence_requests_scoped_by_policy.ordered_by_created_at.map { |d| new_dash_dr[d] }
+  end
+
+  def new_dash_dr
+    @_new_dash_dr ||= DefenceRequest.method(:new).curry(2)[@client]
```
pedrosmmoreira: I'm not entirely sure on this approach. Currying is fine, albeit a "strange" idiom in Ruby. However, here it is used in support of the nested decorator/presenter type object that doesn't seem to be pulling much weight?
I'm OK with going with this approach until we have more info to make a decision on the decorator, but I'd still like to hear more comments from the rest of the team.

app/controllers/dashboards_controller.rb
```diff
@@ -35,4 +36,8 @@ def defence_requests_scoped_by_policy
   def policy_context
     @_policy_context ||= PolicyContext.new(DefenceRequest, current_user)
   end
+
+  def client
+    @_client ||= ServiceRegistry.service(:auth_api_client).new(session[:user_token])
```
robmckinnon: Seems we're not using a @_ prefix convention consistently across controllers. Perhaps we could rename these to `@client` and `@policy_context`?

FriedSock: Happy to do so!

app/models/dashboard.rb
oskarpearson: Is there a reason we're using @_new_dash_dr rather than just @new_dash_dr ? I see we've done it in a few places, but I don't think it's necessary, right? We're not using C :) http://stackoverflow.com/a/25090719

pedrosmmoreira: I believe I am at fault for this. I usually prefix ivars with an underscore to reveal the intent they are not meant to be used in views. I've found it a helpful distinction in controllers and just use elsewhere to be consistent. If the team prefers to lose the prefix, I'm happy with that :)

app/helpers/dashboard_helper.rb
```diff
@@ -6,4 +6,43 @@ def tab_class(type:)
...
+  end
+
+  def arriving_text(dr)
+    if dr.created_at < Time.zone.now.beginning_of_day
+      # before today
+      "Arriving"
+    else
+      if dr.created_at < Time.zone.now.end_of_day
+        # today
+        "Arriving at"
```
robmckinnon: Perhaps it's better to take display text from `en.yml` locales file? e.g. change to `I18n.t('arriving_at')` etc

app/models/dashboard.rb
dp6ai: @pedrosmmoreira i do this as well

PR #169, opened 2015-05-29
Defence request form progressive enhancement - show/hide and disable

robmckinnon:
* Add generic javascript for disabling input when radio checked.
* Add generic javascript for hiding input and showing it when radio checked.
* Set attributes on form elements to enable show/hide and disable javascript where needed.
* Remove old custom javascript for disabling attributes.
* Remove unneeded logic for loading JS partials.
* Javascript unit tests written in jasmine.
* Use darker grey for disabled input background colour.


app/assets/javascripts/application.js
```diff
@@ -13,6 +13,8 @@
 //= require jquery_ujs
 //= require dsds
 //= require date_chooser
+//= require show_hide
+//= require disable_checker
```
novotnyjakub: Should we require these in the files they're needed in? Instead of having all dependencies specified in the main file? I'm just asking, I don't know if there are benefits/requirements of having it here.

app/assets/javascripts/show_hide.coffee
```diff
@@ -0,0 +1,28 @@
...
+# Shows supplied element when radio with ID equal to 'data-show-when' value is checked.
+# Hides supplied element when any other radio in fieldset is checked.
+class ShowHide
+
+  constructor: (elementToShow) ->
+    @elementToShow = elementToShow
+    checkInputId = elementToShow.data("show-when")
+
+    @checkInput = $( "#" + checkInputId )
+    fieldset = @checkInput.closest("fieldset")
```
novotnyjakub: Are we using `fieldset` diligently always around radio buttons?

app/assets/stylesheets/govuk_elements_extensions.scss
```diff
@@ -176,13 +173,14 @@ section {
...
 
+input[disabled],
 input[disabled="disabled"],
 input[readonly="readonly"],
 textarea[disabled],
 textarea[disabled="disabled"],
 textarea[readonly="readonly"] {
   opacity: 1;
-  background-color: $grey-4;
+  background-color: #eee;
```
novotnyjakub: I would maybe abstract this to a colour variable. Just so we have a list of all colours in one place.

app/assets/javascripts/show_hide.coffee
robmckinnon: Good point. I'll add checking fieldset is present in the view spec. For now the contract when using this js is that fieldset is around the radio buttons.

spec/javascripts/disable_checker_spec.coffee
```diff
@@ -0,0 +1,75 @@
+//= require jquery
+//= require jasmine-jquery
+//= require disable_checker
+
+fixtureHtml = (inputId, isChecked) ->
```
novotnyjakub: The `inputId` here is not used.

spec/javascripts/disable_checker_spec.coffee
```diff
@@ -0,0 +1,75 @@
...
+           type="text">
+    <label class="form-checkbox" for="defence_request_detainee_name_not_given">
+      <input name="defence_request[detainee_name_not_given]" value="0" type="hidden">
+      <input value="1" name="defence_request[detainee_name_not_given]" id="defence_request_detainee_name_not_given" type="checkbox" #{checked}>
+      not given</label>
+  </div>
+  </body>
+  """)
+
+fixtureSetup = (element, inputId, context) ->
```
novotnyjakub: Same as above, the `inputId` is not used.

spec/javascripts/disable_checker_spec.coffee
```diff
@@ -0,0 +1,75 @@
...
+  element = null
+
+  afterEach ->
+    element.remove()
+    element = null
+
+  describe "when checkbox not checked", ->
+    beforeEach ->
+      inputId = "defence_request_appropriate_adult_reason"
+      element = fixtureHtml(inputId, false)
```
novotnyjakub: I think it's related to the comments above, but `inputId` is not set anywhere.

spec/javascripts/disable_checker_spec.coffee
```diff
@@ -0,0 +1,75 @@
...
+      fixtureSetup(element, inputId, this)
+
+    describe "after initialization", ->
+      it "leaves input enabled", ->
+        expect(@inputToDisable).not.toBeDisabled()
+
+    describe "radio is checked", ->
```
novotnyjakub: Should this (and some of the following) description be `checkbox` rather than `radio`?

app/assets/javascripts/application.js
robmckinnon: Benefit of having these in the application JS file is that the functionality is available for all views and controllers without further code.

A different approach would be to [put them in a controller specific coffeescript file](http://guides.rubyonrails.org/asset_pipeline.html#controller-specific-assets). However I get the sense these form partials may get rendered via more than one controller.

novotnyjakub: I was more thinking in the `page_initialise` file where they're used. But yeah, no biggie.

spec/javascripts/show_hide_spec.coffee
```diff
@@ -0,0 +1,113 @@
...
+
+  <div data-show-when="defence_request_appropriate_adult_true">
+    <label for="defence_request_appropriate_adult_reason">Reason for appropriate adult</label>
+    <textarea name="defence_request[appropriate_adult_reason]" id="#{ inputId }"></textarea>
+    <input name="defence_request[appropriate_adult_reason_etc]" id="#{ inputId }_etc" type="text"></input>
+  </div>
+  </body>
+  """)
+
+fixtureSetup = (element, inputId, context) ->
```
novotnyjakub: Same as in the other test, `inputId` not being used.

spec/javascripts/disable_checker_spec.coffee
novotnyjakub: Oh, sorry, I see that it's being set here, but it's not used inside the methods, so maybe it can be removed?

spec/views/defence_requests/new.html.erb_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "rails_helper"
```
novotnyjakub: I think it's a good idea to use a view spec to test this. We don't have any view specs though, so I think it should be agreed by the wider team.

app/assets/stylesheets/govuk_elements_extensions.scss
robmckinnon: Ok moving to a colour variable. At present we have not defined any other colour variables so leaving it in the same scss file for now.

app/assets/javascripts/show_hide.coffee
robmckinnon: View spec now checks that radio inputs are enclosed in a fieldset for the show hide related radio inputs.

PR #170, opened 2015-05-29
Single date text field

FriedSock:


app/views/shared/_date_time_form_part.html.erb
```diff
@@ -18,20 +18,11 @@
       <div class="date-field-wrapper">-</div>
       <div class="date-chooser-values">
         <label class="date-chooser-select form-label js-only" data-initial-date="<%= initial_date_type %>">
-          <span class="today" data-date-display="<%= today.to_s(:full) %>" data-day="<%= today.day %>" data-month="<%= today.month %>" data-year="<%= today.year %>"><%= t('today') %></span>
-          <span class="tomorrow" data-date-display="<%= tomorrow.to_s(:full) %>" data-day="<%= tomorrow.day %>" data-month="<%= tomorrow.month %>" data-year="<%= tomorrow.year %>"><%= t('tomorrow') %></span>
+          <span class="today" data-date-display="<%= today.to_s :full %>" data-day="<%= today.day %>" data-month="<%= today.month %>" data-year="<%= today.year %>"><%= t('today') %></span>
+          <span class="tomorrow" data-date-display="<%= tomorrow.to_s :full %>" data-day="<%= tomorrow.day %>" data-month="<%= tomorrow.month %>" data-year="<%= tomorrow.year %>"><%= t('tomorrow') %></span>
```
robmckinnon: Javascript no longer uses `data-day`, `data-month`, and `data-year` attributes. Looks like they are not needed. Suggest we remove them.

app/assets/javascripts/date_chooser.coffee
```diff
@@ -40,16 +38,9 @@ class DateChooser
     selector.html(selector.text())
     @setDate selector
     @enableLink otherSelector
-    @dayInput.focus()
```
robmckinnon: Maybe we could tab focus to the date input after Today/Tomorrow clicked?

We could replace the `@values` variable with a `@dateInput` variable, and use that in the rest of the code, e.g.
```ruby
@dateInput = chooser.find(".date-chooser-values").find(".date")
...
@dateInput.focus()
...
@dateInput.val(date)
```

Gemfile
```diff
@@ -6,6 +6,7 @@ gem "dotenv-rails", "~> 2.0.0", require: "dotenv/rails-now"
 
 gem "audited-activerecord", "~> 4.0.0"
 gem "coffee-rails", "~> 4.1.0"
+gem "chronic"
```
robmckinnon: Do we have to configure chronic to parse dd/dd/dddd as dd/mm/yyyy instead of the US mm/dd/yyyy?

Worth adding a few tests to ensure this works as we want it to.

robmckinnon: Looks like we need to config it. Just did quick test in irb:
```
~ $ irb
irb(main):001:0> require 'chronic'
=> true
irb(main):002:0> Chronic.parse('12/01/2015').month
=> 12
```

PR #171, opened 2015-06-01
Tidy cso form

ansonK:
Story: https://www.pivotaltracker.com/story/show/95707056

Adds missing fields and removes unneeded ones to bring the form in line with designs.

### DIFF
--- a/app/views/defence_requests/show.html.erb
+++ b/app/views/defence_requests/show.html.erb
@@ -14,7 +14,7 @@
 <% end %>
 
 <dl class="labels">
-  <%= display_value "gender", @defence_request.gender %>
+  <%= display_value "gender", t(@defence_request.gender, default: t(:not_given)) %>
   <%= display_value "date_of_birth", date_not_given_formatter(@defence_request, :date_of_birth) %>
   <%= display_value "detainee_address", not_given_formatter(@defence_request, :detainee_address) %>
 </dl>
--- a/app/views/shared/_date_time_form_part.html.erb
+++ b/app/views/shared/_date_time_form_part.html.erb
@@ -1,11 +1,11 @@
 <% today, tomorrow, initial_date, initial_date_type = data_chooser_setup(@defence_request.send(attribute), set_default_date) %>
 <div class="form-group">
   <fieldset class="inline time-select date-chooser">
-    <% if show_label %>
+    <%- if defined?(show_label) && show_label %>
       <label for>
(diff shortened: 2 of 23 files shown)


app/views/shared/_date_time_form_part.html.erb
```diff
@@ -1,11 +1,11 @@
 <% today, tomorrow, initial_date, initial_date_type = data_chooser_setup(@defence_request.send(attribute), set_default_date) %>
 <div class="form-group">
   <fieldset class="inline time-select date-chooser">
-    <% if show_label %>
+    <%- if defined?(show_label) && show_label %>
```
FriedSock: `defined?` smells bad, can we use `local_assigns.fetch :show_label, false` at the top of the file instead? Or just explicitly pass in the variable to each render call.

db/migrate/20150529154920_remove_investigating_officer_shoulder_number.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveInvestigatingOfficerShoulderNumber < ActiveRecord::Migration
+  def change
+    remove_column :defence_requests, :investigating_officer_shoulder_number
+  end
```
FriedSock: We need to add the type of the column here to make the migration reversible.

FriedSock:
LGTM aside from a couple of comments.

app/views/defence_requests/show.html.erb
```diff
@@ -14,7 +14,7 @@
 <% end %>
 
 <dl class="labels">
-  <%= display_value "gender", @defence_request.gender %>
+  <%= display_value "gender", t(@defence_request.gender, default: t(:not_given)) %>
```
novotnyjakub: Just FYI - I've changed the way this part of the detail is rendered - it will use a shared partial. So this will get overriden soon.

ansonK: thats fine

db/migrate/20150529154920_remove_investigating_officer_shoulder_number.rb
ansonK: :+1:

app/views/shared/_date_time_form_part.html.erb
ansonK: local_assigns is a much better way
:+1:

PR #172, opened 2015-06-01
 Dashboard colours

FriedSock:
Need confirmation from kate about what to do for draft requests

### DIFF
--- a/app/assets/stylesheets/dashboard.scss
+++ b/app/assets/stylesheets/dashboard.scss
@@ -1,4 +1,4 @@
-@import "colours";
+@import "colours_extensions";
 @import "elements/elements-typography";
 @import "grid_layout";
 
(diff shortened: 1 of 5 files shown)


ansonK:
Draft requests are not shown on the dashboards

app/assets/stylesheets/dashboard.scss
```diff
@@ -112,11 +112,26 @@ table.dashboard {
...
+    }
+
+    .accepted {
+      border-left: 6px solid #912B88;
+    }
+
+    .completed {
+      border-left: 6px solid #28A197;
+    }
+
```
ansonK: Can we use variables for the colours please

app/assets/stylesheets/dashboard.scss
```diff
@@ -112,11 +112,26 @@ table.dashboard {
     .detainee-container {
       height: 100%;
       margin-left: 1em;
-      border-left: 6px solid #990099;
       padding-right: 0;
       padding: 0px 0 10px 15px;
     }
```
ansonK: Should we keep a default for the state colour?

FriedSock: There shouldn't be anything visible that doesn't have a colour defined, I think it would be misleading to have a default.

PR #174, opened 2015-06-02
Defence request confirmation page

novotnyjakub:
This add a confirmation page after creating a new defence request for CSO. CSO can then change details and eventually `Submit request`. After that the request becomes queued. 

If they decide to abandon defence request before submitting, it stays in the `draft` state. Draft state now doesn't appear on the dashboard and all draft DRs will eventually be cleared after some time (1, 2 days). This has to be implemented later.


app/views/defence_requests/show_draft.html.erb
```diff
@@ -0,0 +1,32 @@
...
+<section class="detention-details">
+  <h2 class="line-above">Detention Details</h2>
+
+  <%= render("defence_requests/show_partials/detention_details") %>
+
+  <%= link_to("Change this", edit_defence_request_path(@defence_request, part: "detention_details"), class: "edit-link") %>
+</section>
+
+<div class="draft-actions">
+    <%= form_tag(transition_defence_request_path(), method: :post) do %>
```
FriedSock: needless parentheses, might see if we can enable the rubocop rule for this.

app/views/defence_requests/show_draft.html.erb
```diff
@@ -0,0 +1,32 @@
...
+
+<section class="detention-details">
+  <h2 class="line-above">Detention Details</h2>
+
+  <%= render("defence_requests/show_partials/detention_details") %>
+
+  <%= link_to("Change this", edit_defence_request_path(@defence_request, part: "detention_details"), class: "edit-link") %>
+</section>
+
+<div class="draft-actions">
```
ansonK: Missing the Custody details section:
Should be:
screen shot 2015-06-02 at 16 55 57

This is how it currently looks:
screen shot 2015-06-02 at 16 56 13

app/views/defence_requests/show_partials/_detention_details.html.erb
```diff
@@ -0,0 +1,12 @@
+<dl class="labels">
+  <%= display_value "time_of_arrest", date_and_time_formatter(@defence_request.time_of_arrest) %>
+  <%= display_value "time_of_arrival", date_and_time_formatter(@defence_request.time_of_arrival) %>
+  <%= display_value "time_of_detention_authorised", date_and_time_formatter(@defence_request.time_of_detention_authorised) %>
```
ansonK: Datetimes should be displayed with the time first:
screen shot 2015-06-02 at 16 59 15

app/views/defence_requests/show_draft.html.erb
novotnyjakub: It's because we don't have that data at the moment. There's no link to the custody suite. I think we should add it then, otherwise it will seem functional to users, but will be fake.

I think when we add the custody suite link, we'll also have to display another tab (or so) on the show page, to keep it consistent.

app/views/defence_requests/show_partials/_detention_details.html.erb
novotnyjakub: Good spot, I'll change that.

app/views/defence_requests/show_draft.html.erb
ansonK: Ok that's cool. 
We can do this in the scope of a separate story as that is a larger chunk of work

PR #176, opened 2015-06-03
[add_build_info_to_container] Add build info to container file

oskarpearson:
Add fields as per https://github.com/ministryofjustice/incident-response-schemas/blob/master/ping-dot-json-schema.json


oskarpearson:
Merging as per https://github.com/ministryofjustice/defence-request-service-auth/pull/90

PR #179, opened 2015-06-03
[add_ping_endpoint] Add ping endpoint as per support requirements

oskarpearson:
Add ping endpoint that returns version information

### DIFF
--- a/lib/ping_response.rb
+++ b/lib/ping_response.rb
@@ -1,0 +1,16 @@
+class PingResponse
+  VERSION_FILE = "/.version"
+
+  UNKNOWN_VERSION_DATA_RESPONSE = {
+    version_number: "unknown",
+    build_date: nil,
+    commit_id: "unknown",
+    build_tag: "unknown"
+  }
+
+  def self.data
+    YAML.load_file(VERSION_FILE)
+  rescue
+    UNKNOWN_VERSION_DATA_RESPONSE
+  end
+end
--- a/lib/ping_response.rb
+++ b/lib/ping_response.rb
@@ -8,9 +8,16 @@ class PingResponse
     build_tag: "unknown"
   }
 
-  def self.data
-    YAML.load_file(VERSION_FILE)
-  rescue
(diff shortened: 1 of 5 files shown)


novotnyjakub:
What is the difference in use of `/status` and `/ping`?

oskarpearson:
@novotnyjakub I meant to write a comment about that.

The ping URL has a very defined format for support team use - https://github.com/ministryofjustice/incident-response-schemas/blob/master/ping-dot-json-schema.json 

Given that the format is fixed, I see the /status URL being used for additional checks down the line. Things like "Is the DB available? Is the redis server/sidekiq instance available" etc etc could be added to /status without breaking the JSON-schema-defined format of ping.json

(There's one concern I have - in the failure case, we currently return a nil for the date-time field defined in that schema - which is a required field.. I'm not sure how we should handle that.)

I'm going to check in with the webops team now and see if the /ping.json response is absolutely set in stone, or if we can add whatever fields we like to it.

lib/ping_response.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  UNKNOWN_VERSION_DATA_RESPONSE = {
+    version_number: "unknown",
+    build_date: nil,
+    commit_id: "unknown",
+    build_tag: "unknown"
+  }
+
+  def self.data
+    YAML.load_file(VERSION_FILE)
```
ansonK: Does the build_date field need to be ISO8601 compliant? 
If so do we rely on the .version file being written correctly, or parse and reformat?

oskarpearson: We are expecting the version file to be written correctly by the build script.

oskarpearson:
After discussion with the web ops team, we need to rename /status to /healthcheck - and have it work similarly to this:

https://github.com/ministryofjustice/peoplefinder/blob/df3aedaaa6ad99dacd244c9b6af26b9a47fdcc93/app/services/health_check_service.rb

I'm going to make a separate pivotal tracker task for this.

ansonK:
:+1:

PR #180, opened 2015-06-03
Reasons for appropriate adult needed and detainee fit for interview

robmckinnon:
* Change appropriate adult reason input to radio buttons.
* Change `appropriate_adult_reason` from text to string.
* Validate `appropriate_adult_reason` is a valid locales key value.
* Implements story: https://www.pivotaltracker.com/story/show/93690592
* Enable javascript for IE 6.
* Tested show/hide javascript on IE 6/7. Worked ok.

### DIFF
--- a/app/helpers/defence_requests_helper.rb
+++ b/app/helpers/defence_requests_helper.rb
@@ -87,4 +87,12 @@ module DefenceRequestsHelper
     end
   end
 
+  def appropriate_adult_reason(defence_request)
+    if defence_request.appropriate_adult?
+      reason = t(defence_request.appropriate_adult_reason).downcase
+      ["–", t("appropriate_adult_reason").downcase, reason].join(" ")
+    else
+      ""
+    end
+  end
 end
--- a/spec/models/defence_request_spec.rb
+++ b/spec/models/defence_request_spec.rb
@@ -17,6 +17,29 @@ RSpec.describe DefenceRequest, type: :model do
         subject.appropriate_adult = true
       end
       it { expect(subject).to validate_presence_of :appropriate_adult_reason }
+
+      context "with invalid reason value" do
+        before do
+          subject.appropriate_adult_reason = "bad value"
+          subject.validate
+        end
+        it { expect(subject.errors[:appropriate_adult_reason].size).to eq(1) }
+      end
+
+      context "with detainee_juvenile value" do
+        before do
+          subject.appropriate_adult_reason = "detainee_juvenile"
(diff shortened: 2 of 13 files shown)


app/helpers/defence_requests_helper.rb
```diff
@@ -87,4 +87,12 @@ def initial_date_type(initial_date, today, tomorrow)
     end
   end
 
+  def appropriate_adult_reason(defence_request)
+    if defence_request.appropriate_adult?
```
novotnyjakub: You don't need to do this check again as the `boolean_with_explanation` helper does it already.

spec/models/defence_request_spec.rb
```diff
@@ -17,6 +17,29 @@
         subject.appropriate_adult = true
       end
       it { expect(subject).to validate_presence_of :appropriate_adult_reason }
+
```
novotnyjakub: Why didn't you use `validate_inclusion_of` Shoulda helper? It should cover all the 3 contexts bellow.

app/helpers/defence_requests_helper.rb
robmckinnon: I recall getting an error in test runs when this check was not present. This helper code is getting executed before the `boolean_with_explanation` helper code runs.

spec/models/defence_request_spec.rb
robmckinnon: Good spot. I'll change this.

app/helpers/defence_requests_helper.rb
novotnyjakub: Good point, I didn't realise that.

novotnyjakub:
I know we haven't worked on the errors yet, but if you don't select this new radio button, you get this:

`Because the detainee is: can't be blank, is not included in the list`.

Should we leave this until further?

robmckinnon:
At present all errors are appearing in the form `label + error message`, e.g.:
```
Detainee name: can't be blank
Address: can't be blank
...
Is an appropriate adult needed?: can't be blank
Is detainee fit for interview: can't be blank
```
Think it's better to make a new story for implementing custom error messages. Expect we'll end up with a generic solution for doing this.

ansonK:
> I know we haven't worked on the errors yet, but if you don't select this new radio button, you get this:
> 
> Because the detainee is: can't be blank, is not included in the list.
> 
> Should we leave this until further?

Yes. We will go over error messaging as a separate piece of work

PR #185, opened 2015-06-04
add status sidebar to show page with dummy content

axemonkey:
Clock icon and full height colour don't work in IE8 or below. These are cosmetic, but if that's a problem other (less elegant) solutions can be found.

### DIFF
--- a/app/presenters/defence_request_presenter.rb
+++ b/app/presenters/defence_request_presenter.rb
@@ -25,24 +25,24 @@ class DefenceRequestPresenter < SimpleDelegator
     end
   end
 
-  def queued_at_time
-    created_at.strftime("%H:%M:%S")
+  def state_updated_time
+    state_updated_at.strftime("%H:%M:%S")
   end
-  alias :submitted_time :queued_at_time
-  alias :accepted_time :queued_at_time
-  alias :completed_time :queued_at_time
-  alias :closed_time :completed_time
-  alias :aborted_time :completed_time
-
-  def last_action_time
-    send "#{state_class}_time"
-  end
+  alias :last_action_time :state_updated_time
 
   private
 
-  attr_reader :client
+  attr_reader :client, :defence_request
+
+  def defence_request
+    __getobj__
+  end
 
   def organisation
     @organisation ||= client.organisation(organisation_uid) if organisation_uid
+  end
+
+  def state_updated_at
+    defence_request.state_updated_at || defence_request.updated_at
   end
 end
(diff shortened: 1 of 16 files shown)


novotnyjakub:
I'm not sure if we should have the sidebar visible with the dummy data. Because we're now user testing with the system, so it can confuse users or bring up unnecessary concerns about functionality.

axemonkey:
Sure. Kate said to stick dummy data in, but you're right - if we're testing then this should be either parked or updated to use real data. I'll close the PR for now.

ansonK:
@novotnyjakub agreed. Lets park the PR for a rails dev to pick up and make dynamic

novotnyjakub:
I think it depends on how we generally want to handle the frontend work. If we want to merge HTML + CSS in and have it ready for further implementation. Or we pick up the implementation straight away. 

We could easily loose track and have these PRs hanging around for quite a while.

ansonK:
We can link the branch in trello - that way it is kept track of and can be reassigned etc

app/presenters/defence_request_presenter.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def state_updated_time
+    state_updated_at.strftime("%H:%M:%S")
+  end
+  alias :last_action_time :state_updated_time
+
+  private
+
+  attr_reader :client, :defence_request
+
```
FriedSock: The `attr_reader :defence_request` can be removed here right?

spec/models/defence_request_spec.rb
```diff
@@ -1,6 +1,24 @@
 require "rails_helper"
 
 RSpec.describe DefenceRequest, type: :model do
+  describe "state_updated_at" do
+    subject { FactoryGirl.create :defence_request }
+
+    it "is set equal created_at when creating" do
+      expect(subject.state_updated_at).to be_within(1.second).of(subject.created_at)
```
FriedSock: don't like it, can we stub time or something instead of assuming that these tests will run quickly?

app/presenters/defence_request_presenter.rb
ansonK: yes that can be removed. Good catch

spec/models/defence_request_spec.rb
ansonK: This value is set within the create transaction if it is > 1sec different it is an indication that something is wrong, not just "slow" tests

PR #186, opened 2015-06-04
Update cso form and details view with content tweaks

ansonK:
- Reorders input elements on CSO form
- Updates label translations
- Changes "Not specified" gender option to "Unspecified"
- Adds ability to display help text for a specific date field

### DIFF
--- a/app/views/dashboards/_dashboard.html.erb
+++ b/app/views/dashboards/_dashboard.html.erb
@@ -4,8 +4,6 @@
 
 <%= flash_messages %>
 
-<h1></h1>
-
 <% if check_policy_clause(@policy, :create?) %>
   <div class="new_defence_request">
     <%= link_to t("new_request"), new_defence_request_path, role: "button", class: "button" %>
--- a/app/views/dashboards/_dashboard.html.erb
+++ b/app/views/dashboards/_dashboard.html.erb
@@ -4,7 +4,7 @@
 
 <%= flash_messages %>
 
-<h1><%= "#{t(@dashboard.user_role)}  #{t('dashboard')}" %></h1>
+<h1></h1>
 
 <% if check_policy_clause(@policy, :create?) %>
   <div class="new_defence_request">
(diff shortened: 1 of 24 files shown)


app/views/defence_requests/form_partials/_detainee_details.html.erb
```diff
@@ -24,8 +24,8 @@
       <%= t('transgender') %>
     </label>
     <label class="block-label form-label" for="defence_request_gender_not_specified">
```
novotnyjakub: This label is now for a wrong field. It should be `..._gender_unspecified`. Clicking on the label doesn't work :) Maybe we should always use the `f.label` helper?

app/views/dashboards/_dashboard.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <%= flash_messages %>
 
-<h1><%= "#{t(@dashboard.user_role)}  #{t('dashboard')}" %></h1>
+<h1></h1>
```
novotnyjakub: We probably shouldn't have empty `<h1>`? We should either remove it (which I'm not sure how is it with accessibility) or put something in.

ansonK: yeah good spot

app/views/defence_requests/show.html.erb
```diff
@@ -31,7 +31,7 @@
       <%= render "defence_requests/show_partials/interview_tab" %>
     </div>
     <div id="case-details">
-      <h3 class="nonjs"><%= t("detention_details")%></h3>
+      <h3 class="nonjs"><%= t("case_details")%></h3>
```
novotnyjakub: So is `Case Details` the final copy? Because the latest design was using `Detention Details` and I've put it in a few places in the code.

app/views/shared/_date_time_form_part.html.erb
```diff
@@ -4,6 +4,11 @@
     <%- if local_assigns.fetch(:show_label, false) %>
       <label for>
```
novotnyjakub: I know this wasn't change, but I just noticed that we have `label` inside `label`.

app/views/defence_requests/show.html.erb
ansonK: Case details for CSO's
Detention details for Solicitors

(same thing, different terminology for different user groups)

novotnyjakub: Damn, that's gonna be interesting to solve.

dp6ai: This seems like a terrible idea. If one user talks to another having different terminologies will be really confusing...

PR #187, opened 2015-06-04
Update interview start time now accepts the correct parameters

ansonK:
Fixes bug where changing the interview time was not working as the strong params in the controller where not allowing the correct values through.

Also added a feature spec as there was none.


ansonK:
Merging as this is needed for user testing first thing tomorrow

PR #189, opened 2015-06-05
The day after tomorrow

FriedSock:
The end is nigh

### DIFF
--- a/app/models/day.rb
+++ b/app/models/day.rb
@@ -10,4 +10,5 @@ class Day
   YESTERDAY = new "yesterday", Date.yesterday
   TODAY = new "today", Date.today
   TOMORROW = new "tomorrow", Date.tomorrow
+  DAY_AFTER_TOMORROW = new "day_after_tomorrow", Date.tomorrow + 1
 end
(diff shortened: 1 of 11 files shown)


app/models/day.rb
```diff
@@ -0,0 +1,14 @@
+class Day
```
novotnyjakub: There's no spec for this model. I know it may seem silly, but I feel we should have models covered.

novotnyjakub:
Generally speaking I'm not sure about the design. It's quite squeezed in together when there are 3 options and one of them is quite long.

field

app/models/day.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def initialize(name, date)
+    @name = name
+    @date = date
+  end
+
+  YESTERDAY = new "yesterday", Date.yesterday
+  TODAY = new "today", Date.today
+  TOMORROW = new "tomorrow", Date.tomorrow
+  DAY_AFTER_TOMORROW = new "day_after_tomorrow", Date.tomorrow + 1
```
ansonK: What happens when the rails process runs for longer than 1 day?
Because of the way the Date instances are cached as constants they would not update...

FriedSock:
@novotnyjakub re: design. Lets see how it tests!

PR #191, opened 2015-06-05
Edit defence request from detail

novotnyjakub:
The Submitted defence request show page now has 2 links to "Change this". One for the detainee details and one in the Case details tab.


ansonK:
After updating Case details the page reloads showing the Interview tab.
Should it not reload with the Case details tab showing (as that was where the User was before) ?

novotnyjakub:
Good point, I'll take a look at that.

novotnyjakub:
OK @ansonK, this should be ready to go. I've added the support for coming back to the same tab. I've also refactored a bit some views and removed the `defence_request_management_spec` for good.

app/controllers/defence_requests_controller.rb
```diff
@@ -1,6 +1,7 @@
 class DefenceRequestsController < BaseController
 
   include DefenceRequestConcern
+  include DefenceRequestsHelper
```
ansonK: Do we need to include a view helper into the controller?

novotnyjakub: Yeah, so this is something I'm not sure about. Essentially `ApplicationHelper` is automatically available everywhere and I thought we're just splitting it semantically to smaller focused helpers. But if the purpose of all other helpers apart from `ApplicationHelper` is just views, than I'll move my url helper there.

ansonK:
:+1:

PR #192, opened 2015-06-05
Fix tabs to float correctly in low IEs

axemonkey:
Tested and now floating correctly in IE6-10, and still working fine in Chrome and Firefox

### DIFF
--- a/app/assets/stylesheets/tabs.scss
+++ b/app/assets/stylesheets/tabs.scss
@@ -69,6 +69,9 @@ ul.tabs {
   padding: 0;
 
   li {
+    width: 33%; // default for low IEs
+    display: inline-block;
+
     @include media(mobile) {
       @include grid-column(1 / 2, $full_width: mobile);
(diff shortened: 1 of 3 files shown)


axemonkey:
Hold it - just realised there are tabs on more than one page, and they don't work correctly on the others for some reason. Don't merge yet.

novotnyjakub:
The tabs seem to behave differently on the defence request detail page and the dashboard. On the detail page, the tab link stays outlined after clicking on it.

novotnyjakub:
Also I just noticed that the side bar is not there, so maybe you should rebase on master, to make sure nothing gets lost?

axemonkey:
Ah yes. One is JS-powered and one is a full page refresh

novotnyjakub:
I understand that, but it has an orange border when selected. I guess it's the :active state, but it looks weird. If you think it's fine, I won't block it :)

PR #193, opened 2015-06-05
Error summary links and embedded validation messages

robmckinnon:
* Create a custom form builder to place error messages next to form labels.
* Use new form builder helpers in the defence request form(s).
* Style validation errors using new GOV.UK elements styles.
* WIP, needs a rebase, and probably a refactoring before ready.
* Implements: https://trello.com/c/pVQjCtQK/5-3-as-a-cso-when-i-am-filling-changing-a-defence-request-i-d-like-to-see-errors-on-the-form and https://trello.com/c/tkhYzG3U/6-5-form-errors-have-link-to-anchor-points-within-form


FriedSock:
Seems to work, although the highlighting of parts of date / datetime fields is a bit hit and miss:

screen shot 2015-06-10 at 11 57 26


screen shot 2015-06-10 at 12 23 10

The highlighting is not based on whether or not there is a problem with the individual parts of the date, we could do a chunk of work to make that happen: or highlight all the text fields, or none of them.

ansonK:
The error message should be displayed on each field between the label and the input elements:

screen shot 2015-06-10 at 16 35 11

ansonK:
Interview start time (using the datetime component) is missing the optional text in the label:

screen shot 2015-06-10 at 16 35 59

it should look like:
screen shot 2015-06-10 at 16 36 14

app/builders/labelling_form_builder.rb
```diff
@@ -0,0 +1,75 @@
...
+      unless error_message.blank?
+        f.content_tag(:span, class: "error-message") do
+          error_message
+        end
+      end
+    end
+
+    def parent_id
+      f.object_name.to_s.tr("[]","_").squeeze("_")
+    end
```
ansonK: This seems more like a base class for each 'field' class rather than an object to be composed with.
Can we change this to be an abstract base class which each of the implementing classes (like TextField) extend form?

app/builders/labelling_form_builder.rb
```diff
@@ -0,0 +1,75 @@
...
+    content = block_given? ? capture(&block) : ""
+    TextField.new(self, a, options).content content
+  end
+
+  # Defaults to "Yes" "No" labels on radio inputs
+  #TODO: make legend do something
+  def radio_button_fieldset(attribute, legend, options={})
+    a = Attribute.new(self, attribute)
+    RadioButtonFieldset.new(self, a, options).content
+  end
```
ansonK: Would be good to have a consistent method names for the API to the builder - eg 
- text_field_input and radio_button_input OR
- text_field and radio_button(s)

app/builders/labelling_form_builder.rb
```diff
@@ -0,0 +1,75 @@
+class LabellingFormBuilder < ActionView::Helpers::FormBuilder
+  include ActionView::Helpers::CaptureHelper
+  include ActionView::Helpers::TagHelper
+  include ActionView::Context
+
+  def form_group(attribute, options={}, &block)
+    a = Attribute.new(self, attribute)
+    FormGroup.new(self, a, options).content capture(&block)
+  end
```
ansonK: As all fields need to be wrapped by a form-group element this can be moved into the base class that TextField extends

app/views/defence_requests/form_partials/_case.html.erb
```diff
@@ -1,42 +1,46 @@
...
-  </div>
-
-  <%= render('shared/date_time_form_part',
-             attribute: :time_of_arrest,
-             show_label: true,
-             dates: [Day::YESTERDAY, Day::TODAY],
-             initial_date_index: 1,
-             f: f) %>
+  <%= f.form_group(:time_of_arrest) do %>
+    <% render('shared/date_time_form_part',
```
novotnyjakub: Here we combine the form builder and a partial, which then uses some parts of the builder. I'm just thinking, that it would be nice if we could just use the builder as all the components should be pre-defined and then we can just compose all forms from it. But maybe that's too complicated?

config/locales/en.yml
```diff
@@ -1,6 +1,14 @@
 en:
+  defence_request:
+    appropriate_adult_reason:
+      label: Because the detainee is
+    unfit_for_interview_reason:
+      label: Reasons
```
novotnyjakub: Aren't these duplicates? These are already on line `233` and `234`.

novotnyjakub:
There's no spec for most of the builder classes. Should we have them?

app/builders/labelling_form_builder.rb
robmckinnon: If renaming methods, probably best to avoid names clashing with existing form helper methods, see list here: http://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html

e.g. Rails already defines `text_field` and `radio_button` methods.

PR #198, opened 2015-06-10
Reorganise CSS files

axemonkey:


chriscarter90:
LGTM :shipit: Organisation of chaos ftw. :+1:

novotnyjakub:
I would just suggest to rename the `govuk_elements_extensions` to something else, or even just put stuff to `application.css.scss`. Because was quite confusing me at the beginning and the content is not really direct extensions, we're using govuk stuff, but it's all application specific.

axemonkey:
Can't put everything into `application.css.scss` as then the conditional comments in the layout won't be able to pull in the correct sass-ified version of the govuk stuff. But I'll rename to something generic like `main.scss` if that suits better.

PR #199, opened 2015-06-10
[force_ssl_bugfix] Corrected the setting name for force_ssl

oskarpearson:
The environment variable wasn't being honored by rails, and was defaulting to false.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -43,7 +43,7 @@ Rails.application.configure do
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = Settings.force_ssl
+  config.force_ssl = Settings.rails.force_ssl
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.


config/environments/production.rb
```diff
@@ -43,7 +43,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = Settings.force_ssl
+  config.force_ssl = Settings.rails.force_ssl
```
oskarpearson: see https://github.com/ministryofjustice/defence-request-service/blob/master/config/settings.yml#L6-L8 - where the force_ssl item is within the 'rails:' section

.env.development
```diff
@@ -1,11 +1,13 @@
 # App settings
-export DISABLE_DASHBOARD_REFRESH=true
+DISABLE_DASHBOARD_REFRESH=true
 
 # Database settings
 DEFENCE_SOLICITOR_DATABASE_HOST=localhost
 DEFENCE_SOLICITOR_DATABASE_PASSWORD=""
 DEFENCE_SOLICITOR_DATABASE_USERNAME="postgres"
 
+FORCE_SSL=false
```
oskarpearson: Question for everyone - given that the force_ssl option is only set in production.rb, do you think this is overkill?

oskarpearson: https://github.com/ministryofjustice/defence-request-service/blob/master/config/environments/production.rb#L46 but not in https://github.com/ministryofjustice/defence-request-service/blob/master/config/environments/development.rb

oskarpearson:
Answering my own questions above, I'm not going to set the variable for other environments. It's unnecessarily and breaks the principle of least surprise ("I changed this variable, but it doesn't have the impact I expected")

novotnyjakub:
Can we set it false for `development` and `test` environments? 

I'm suspecting that may be the reason the build failed as well.

oskarpearson:
Chatted to Jakub - I've not added to the development/test environments because only production.rb uses the environment variable

PR #201, opened 2015-06-11
tidy up broken field alignments in IE6/7

axemonkey:
IE6 before: 
ie6before

IE6 after: 
ie6after

IE7 before: 
ie7before

IE7 after: 
ie7after


ansonK:
:+1:

PR #203, opened 2015-06-12
CSO can update interview time from the defence request show page

novotnyjakub:


app/views/interview_start_times/_form.html.erb
```diff
@@ -0,0 +1,31 @@
...
+     url: defence_request_interview_start_time_path(defence_request_id: @defence_request.id),
+     method: :patch,
+     builder: LabellingFormBuilder,
+     as: :defence_request
+   }
+   remote = true unless local_assigns.has_key?(:remote)
+   if remote
+     form_params.merge!(remote: true, data: { container: "#interview" }, authenticity_token: true)
+   end
+%>
```
ansonK: Were we going to clean this up? Either into a helper or decorator?

novotnyjakub: Oh, damn, I totally forgot that. Yes, I'll clean that up.

novotnyjakub: This is now in a helper.

PR #210, opened 2015-06-17
Error validation

FriedSock:
Updated copy for error messages

### DIFF
--- a/app/fields/date_field.rb
+++ b/app/fields/date_field.rb
@@ -7,7 +7,7 @@ class DateField
     begin
       Date.new year.to_i, month.to_i, day.to_i
     rescue => e
-      errors.add :base, "Invalid Date"
+      errors.add :base, error_message_lookup_proc[:invalid]
     end
   end
 
@@ -33,6 +33,16 @@ class DateField
         dob.year = date.year
       end
     end
+  end
+
+  def set_error_message_lookup_proc!(error_proc)
+    @error_message_lookup_proc = error_proc
+  end
+
+  private
+
+  def error_message_lookup_proc
+    @error_message_lookup_proc || ->(a){ a }
   end
 
 end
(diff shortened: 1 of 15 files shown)


ansonK:
Error message when Appropriate adult is chosen but no sub-option is selection has the default error text appended when it shouldn't?

screen shot 2015-06-17 at 17 48 42

app/fields/date_field.rb
```diff
@@ -35,4 +35,14 @@ def self.from_persisted_value(date)
...
   end
 
+  def set_error_message_lookup_proc!(error_proc)
+    @error_message_lookup_proc = error_proc
+  end
+
+  private
+
+  def error_message_lookup_proc
+    @error_message_lookup_proc || ->(a){ a }
```
novotnyjakub: Do you want to cache the lookup proc with `||=`? Or is it on purpose always creating a new one.

FriedSock: It was to provide a fallback if the proc had not been set, I guess there is no harm in using `||=` as `set_error_message_lookup_proc!` should still work

novotnyjakub: It will work, I just wanted to check what the intention is. I was just thinking that if this method gets called 20 times for the form, we could cache it. But mostly I guess it'll be set explicitly by the setter..

spec/features/solicitor/defence_request_management_spec.rb
```diff
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
```
novotnyjakub: I'm sure this doesn't belong here :-)

app/fields/date_field.rb
FriedSock: It should only get called once, when trying to submit the form, but it is a valid optimisation nonetheless.

PR #212, opened 2015-06-18
Update omniauth dsds to include organisations

novotnyjakub:
This allows to use the `master` of the `Auth` app with the latest changes to the way permissions and roles are handled. 

Changes:

* the service app now decorates the `Omniauth::Dsds::User` with it's own `ServiceUser`, which selects one organisation to work with (at the moment only `custody_suite` type is allowed

* I commented out `cco` and `solicitor` feature tests, so we can go through them again and remove the functionality as well as the tests 
* I removed the `cco` and `solicitor` policies


ansonK:
lgtm

PR #214, opened 2015-06-19
Fix for passing invalid tab param to show page

FriedSock:


app/controllers/defence_requests_controller.rb
```diff
@@ -10,7 +10,7 @@ class DefenceRequestsController < BaseController
   helper_method :defence_request_path_with_tab
 
   def show
-    @tab = params[:tab]
+    @tab = params[:tab] && %(interview case-details).include?(params[:tab]) ?  params[:tab] : nil
```
chriscarter90: Should this, instead of returning nil, just return the first/default?

app/views/defence_requests/show.html.erb
```diff
@@ -17,7 +17,7 @@
 <%= content_tag("div", class: "js-tabs cf", id: "edit-tabs", data: { activatefirst: @tab.blank? ? "true" : "false" }) do %>
   <div class="tabs-nav">
     <ul class="cf tabs js-tabs-nav js-only">
-      <li>
+      <li <%= tab_active_class(@tab == "intervew") %>>
```
chriscarter90: typo on `interview`

spec/features/cso/viewing_defence_request_spec.rb
```diff
@@ -37,4 +37,17 @@
...
+    cso_user = create :cso_user
+    defence_request = create :defence_request, :queued
+
+    login_with cso_user
+    within "tr#defence_request_#{defence_request.id} td.actions" do
+      find("a").click
+    end
+
+    visit defence_request_path(defence_request, tab: "googly woogly wooo")
+    expect(find(:css, ".tabs").first("li")["class"]).to include "is-active"
```
chriscarter90: How does this test pass given the typo above?

app/controllers/defence_requests_controller.rb
ansonK: Also whenever params[:tab] is checked the value must be whitelisted - not like https://github.com/ministryofjustice/defence-request-service/pull/214/files#diff-5e2ba0f9f4b5b18a5e05aebd97940c33R45 where the value is directly passed to a route helper (security risk)

spec/features/cso/viewing_defence_request_spec.rb
FriedSock: What typo?

app/controllers/defence_requests_controller.rb
FriedSock: The tab javascript we are using has an `activatefirst` data attribute which we are defaulting to when to tab has been specified.

FriedSock:
Maybe we could standardise what parameter to use for selecting tabs. I think `:tab` is probably better to use that `:part` for edit and update actions. And it would make the code slightly simpler

ansonK:
lgtm

novotnyjakub:
I thought about it as well, but the semantics is different. It's a `tab` on the `show` page, but it's really a form `part` on the `edit` actions. But I'm not too fuzz about it, if everybody agrees on `tab`, I'll change it.

ansonK:
@novotnyjakub lets leave it for the time being

PR #215, opened 2015-06-19
Assign custody suite to defence request

novotnyjakub:
I also removed unused policies and related code. 

There will be one more PR for this user story, which improves the `create?` policy check, which should make sure that cso can't create a defence request for any other custody suite.


app/controllers/defence_requests_controller.rb
```diff
@@ -35,19 +35,11 @@ def edit
...
   def update
-    if update_and_accept?
-      if @defence_request_form.submit(defence_request_params) && accept_and_save_defence_request
-        redirect_to(dashboard_path, notice: flash_message(:updated_and_accepted, DefenceRequest))
-      else
-        redirect_to(edit_defence_request_path, alert: flash_message(:failed_to_update_not_accepted, DefenceRequest))
-      end
+    @part = params[:part]
+    if @defence_request_form.submit(defence_request_params)
+      redirect_to(defence_request_path_with_tab(@part), notice: flash_message(:update, DefenceRequest))
```
ansonK: Whenever params[:tab] is checked the value must be whitelisted -  the value is directly passed to a route helper (security risk)

novotnyjakub: I haven't actually changed this, I just deleted the unused condition branch in the method.

I think this will (should) get fixed when the "tab bug" is fixed.

PR #216, opened 2015-06-22
gif workaround for IE6

axemonkey:
It's one extra server request, but it fixes the look of tabs in IE6.


chriscarter90:
Not quite a 1px gif but it'll do. :shipit:

PR #217, opened 2015-06-22
Fix notification dl styling

axemonkey:
removes the need to split DLs when doing notifications


novotnyjakub:
Does this apply only to the static markup for firm admin at the moment?

ansonK:
@novotnyjakub it an example of how to display notifications using that design so can apply to other areas, but solicitor admin first

PR #218, opened 2015-06-22
Improve policy for creating / editing requests by cso

novotnyjakub:
The policy now checks the action with the changed object. So a cso user can't assign (re-assign) the defence request to another custody suite. 

I've done a bit of a cleanup, but I'd like to do more. In `DefenceRequestController` I've removed one of the `before_action` filters, which was creating the form. But I would like to take it further and remove at least the `find_defence_request` filter and maybe even the authorisation one. Because it's now very unclear which instance of `DefenceRequest` is actually assigned to `@defence_request`. For example for `edit` and `update`, we actually use `DefenceRequestPresenter`. Which is not right.


ansonK:
Looks good.

PR #220, opened 2015-06-23
notification fixing

axemonkey:
* de-clutter markup for notifications in DL lists
* looks like designs in good browsers and IE7+
* looks acceptable in IE6

(This is part 2 of PR #217 which I should have marked as WIP)


chriscarter90:
:+1: :shipit:

PR #221, opened 2015-06-23
Namespace views

FriedSock:
Namespace any cso related activity to "Custody Suite"


FriedSock:
Should we delete any functionality regarding notifications? A bunch of the tests had to be made pending.

novotnyjakub:
@FriedSock I think we should as notifications have not yet been designed.

app/controllers/sessions_controller.rb
```diff
@@ -4,11 +4,20 @@ class SessionsController < ApplicationController
...
 
   private
 
+  def redirect_url
+    custody_suite_root_url if user_role == "cso"
+  end
+
+  # TODO: This is hacky
+  def user_role
+    @user_role ||= current_user.organisations.map { |o| o["roles"] }.flatten.uniq.first
```
novotnyjakub: There's `current_user.roles` array now, which should be used. We only use roles from 1 organisation at the time.

Also I'm thinking that as we have a very similar code in `PolicyContext`, we should maybe put this on the current user, so at least it's in 1 place and when we change it we have it under control. 

If we don't consolidate, this should be: `current_user.roles.uniq.first`.

app/models/defence_request.rb
```diff
@@ -119,8 +115,4 @@ def closed?
   def notify_interview_start_change
```
novotnyjakub: I would suggest to remove this as well as you're on it.. Cause solicitors won't have access until much later. What do you think @ansonK ?

config/routes.rb
```diff
@@ -1,18 +1,18 @@
...
+      get :closed, action: :show, id: :closed
+      get :refresh_dashboard
+    end
+
+    resources :defence_requests, except: [:index] do
+      resource :interview_start_time, only: [:edit, :update]
     end
+
+    root controller: :dashboards, action: :show
   end
```
novotnyjakub: This now looks great, much better!

spec/models/defence_request_spec.rb
```diff
@@ -89,7 +89,7 @@
     end
 
     shared_examples "transition possible" do |event|
-      specify { expect{ subject.send(event) }.to_not raise_error }
+      xspecify { expect{ subject.send(event) }.to_not raise_error }
```
novotnyjakub: Why is this marked as pending? This seems like an important behaviour.

spec/models/defence_request_spec.rb
```diff
@@ -153,7 +153,7 @@
         @dr_with_dscc = FactoryGirl.create(:defence_request, :with_dscc_number, :acknowledged)
       end
 
-      it "notifies the solicitor"  do
+      xit "notifies the solicitor"  do
```
novotnyjakub: As above, I'd delete the solicitor notification code altogether.

app/presenters/firm_admin/defence_request_presenter.rb
```diff
@@ -0,0 +1,48 @@
+class FirmAdmin::DefenceRequestPresenter < SimpleDelegator
```
novotnyjakub: This is at the moment just C&P of the `DefenceRequestPresenter`. If we don't need it different from the other one, than I'd delete it and even for firm admin use the original one.

novotnyjakub: It actually seems it's not used.

FriedSock: Damn, staged these by accident, good catch!

app/models/defence_request.rb
ansonK: yup remove it

app/controllers/sessions_controller.rb
FriedSock: I don't think we should consolidate this yet, I think it would be weird to have a `role` and a `roles` method on the user. And there may still be benefit to seeing what other roles the user has.

novotnyjakub: Ok, agree. But can you please use `current_user.roles.uniq.first` instead then? Because if you map the organisations like that, you may pick a role from a different organisation, than the `current valid` one.

PR #222, opened 2015-06-23
[asset_content_encoding] Revert to known-good fog-aws version

oskarpearson:
The asset_sync gem uses fog-aws to upload assets to S3. However fog-aws > 0.1.2 sets the Content-Encoding to null value.

This gem can be removed once https://github.com/fog/fog-aws/issues/130 is correctly resolved


novotnyjakub:
It's 3 minor versions downgrade, aren't we loosing any other features?

oskarpearson:
There are indeed a lot of changes between those versions (https://github.com/fog/fog-aws/compare/v0.1.2...v0.5.0 shows them). However, pushing the assets to S3 (our only current usecase) works with either version.

My vote is we keep an eye on that bug and if things are resolved we can upgrade. Hopefully doesn't take too long.

PR #225, opened 2015-06-25
Make the address a textarea

novotnyjakub:
This piece of work discovered quite a big bug when editing a request - checkboxes and radios are not pre-populated. 

But I'll tackle it separately.

### DIFF
--- a/app/builders/text_field.rb
+++ b/app/builders/text_field.rb
@@ -3,6 +3,7 @@ class TextField < FormGroup
   def initialize(form, attribute, options)
     @input_data = options.fetch :input_data, nil
     @input_class = options.fetch :input_class, nil
+    @multiline = options.fetch :multiline, false
     super
   end
 
@@ -25,7 +26,11 @@ class TextField < FormGroup
   end
 
   def value
-    f.text_field attribute, input_options
+    if @multiline
+      f.text_area attribute, input_options.merge({rows: @multiline})
+    else
+      f.text_field attribute, input_options
+    end
   end
 
   def max_length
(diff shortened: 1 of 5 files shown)


app/builders/text_field.rb
```diff
@@ -25,7 +26,11 @@ def labelled_text_field
   end
 
   def value
-    f.text_field attribute, input_options
+    if @multiline
+      f.text_area attribute, input_options.merge({rows: @multiline})
+    else
+      f.text_field attribute, input_options
+    end
```
ansonK: This could be a text_area class the extends text_field.
That way the form can expose a f.text_area_input helper method that is more descriptive than passing in a :multiline param to a text_area helper that then outputs a text area

app/views/custody_suite/defence_requests/form_partials/_detainee_details.html.erb
```diff
@@ -14,6 +14,7 @@
 <% end %>
 
 <%= f.text_field_input(:detainee_address,
+    multiline: 5,
```
ansonK: Related to the above comment having a text_area_input helper would be cleaner and better revealing about what the output from the helper method is going to be

app/builders/text_field.rb
novotnyjakub: I didn't want to do it, because of the already large complexity of the custom form builder. But yeah, I can spend some more time on this.

app/views/custody_suite/defence_requests/form_partials/_detainee_details.html.erb
novotnyjakub: I actually did find this more accurate, because from the abstraction perspective, we don't mind which html element is going to be rendered, just that it has multiple lines. 
But I'll change this.

novotnyjakub:
Ok @ansonK this should now be ready to merge.

PR #226, opened 2015-06-26
Delegate attributes directly to DefenceRequest from DefenceRequestForm

novotnyjakub:
This fixes the problem with checkboxes and radio buttons not being pre-selected when editing the defence request. 

@FriedSock pointed out that if we had a field which expects a numeric value, this could cause it to save incorrectly. But I didn't seem to reproduce it on the existing form, so I think we can go with this solution for now.


chriscarter90:
LGTM :+1: :shipit:

